PR #5, opened 2017-10-13
Adding authoring back button

ekosmin:
Hi @pjanik, I think this takes care of the back button! Seems to pass `lint` as well. Let me know what you think - maybe we can take a look at it together tomorrow after standup.

### DIFF
--- a/js/components/authoring.js
+++ b/js/components/authoring.js
@@ -21,11 +21,12 @@ export default class Authoring extends PureComponent {
     this.state = {
       step: 1
     }
-    this.handleButtonClick = this.handleButtonClick.bind(this)
+    this.handleNextButtonClick = this.handleNextButtonClick.bind(this)
+    this.handleBackButtonClick = this.handleBackButtonClick.bind(this)
     this.handleInteractionChange = this.handleInteractionChange.bind(this)
   }
 
-  get buttonLabel () {
+  get nextButtonLabel () {
     const { step } = this.state
     return step === 3 ? 'finish' : 'next'
   }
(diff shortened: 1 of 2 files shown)


js/components/authoring.js
```diff
@@ -54,6 +55,18 @@ export default class Authoring extends PureComponent {
...
 
+  handleBackButtonClick () {
+    const { step } = this.state
+    const { setOption } = this.props
+    if (step === 2) {
+      this.unloadModel()
+    } else if (step === 3) {
+      setOption('interaction', 'drawContinent')
+      setOption('selectableInteractions', [ 'drawContinent', 'eraseContinent', 'none' ])
+      this.setState({step: 2})
```
pjanik: I guess somewhere else in the code, we call the same set of functions to change step from 1 to 2. We could extract that to something like `setStep2()`, so we don't end up with different UI in the future after going next and back (if you forget to modify one of those sets). I might apply to a few lines in `unloadModel` too.

js/components/authoring.js
```diff
@@ -54,14 +55,36 @@ export default class Authoring extends PureComponent {
...
+
+  unloadModel () {
+    const { unloadModel, setOption } = this.props
+    unloadModel()
+    setOption('interaction', 'none')
+    setOption('selectableInteractions', [])
+    this.setState({step: 1})
+  }
+
+  setStep2 () {
```
ekosmin: I refactored the common code into a single method here! It wasn't as straightforward to pull out a method for setting step 1, since most options are set from initialization the first time it's loaded. So, I left `unloadModel` alone.

PR #6, opened 2017-10-19
Back button stores snapshots between steps

ekosmin:
Previously, going back from the force step to the drawing step would not remove added forces. Now, a snapshot is stored whenever the "next" button is pressed, and one is restored whenever the "back" button is pressed. This will make "undoing" steps like the force drawing simpler in the future.

### DIFF
--- a/js/components/authoring.js
+++ b/js/components/authoring.js
@@ -39,12 +39,12 @@ export default class Authoring extends PureComponent {
 
   handleNextButtonClick () {
     const { step } = this.state
-    const { setOption } = this.props
+    const { setOption, takeSnapshot } = this.props
     if (step === 2) {
-      setOption('interaction', 'force')
-      setOption('selectableInteractions', [ 'force', 'none' ])
-      this.setState({step: 3})
+      takeSnapshot()
+      this.setStep3()
     } else if (step === 3) {
+      takeSnapshot()
       setOption('interaction', 'none')
       setOption('selectableInteractions', [])
       setOption('colormap', 'plate')
@@ -63,10 +63,15 @@ export default class Authoring extends PureComponent {
 
   handleBackButtonClick () {
     const { step } = this.state
+    const { restoreSnapshot } = this.props
     if (step === 2) {
       this.unloadModel()
     } else if (step === 3) {
+      restoreSnapshot()
       this.setStep2()
+    } else if (step === 4) {
+      restoreSnapshot()
+      this.setStep3()
     }
   }
 
@@ -94,6 +99,14 @@ export default class Authoring extends PureComponent {
     setOption('interaction', 'drawContinent')
     setOption('selectableInteractions', [ 'drawContinent', 'eraseContinent', 'none' ])
     this.setState({step: 2})
+  }
+
+  setStep3 () {
+    const { setOption } = this.props
+    setOption('interaction', 'force')
+    setOption('selectableInteractions', [ 'force', 'none' ])
+    setOption('colormap', 'topo')
+    this.setState({step: 3})
   }
 
   renderPreset (presetInfo) {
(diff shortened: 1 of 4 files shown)


js/plates-model/model-worker.js
```diff
@@ -23,12 +23,10 @@ function workerFunction () {
   if (props.playing) {
     if (config.snapshotInterval && model.stepIdx % config.snapshotInterval === 0) {
       if (model.stepIdx === 0) {
+        snapshots.length = 0
```
ekosmin: I know I'm complicating snapshots a bit by using them for saving state during authoring. But, the two changes I made (clear snapshots at the beginning of a run, and never "double back" snapshots if you're at the beginning of a run) seemed reasonable enough, so I left it this way.

pjanik: I see. It took me a while to understand how it all works with authoring, but yeah, it makes sense. I'd just add a comment here that `snapshots.length = 0` is cleaning up authoring snapshots.

js/plates-model/model-worker.js
```diff
@@ -80,13 +85,15 @@ onmessage = function modelWorkerMsgHandler (event) {
...
+  } else if (data.type === 'takeSnapshot') {
+    takeSnapshot()
   } else if (data.type === 'restoreSnapshot') {
     let serializedModel
     if (snapshots.length === 0) {
       serializedModel = initialSnapshot
     } else {
       serializedModel = snapshots.pop()
-      if (snapshots.length > 0 && model.stepIdx < serializedModel.stepIdx + 20) {
+      if (snapshots.length > 0 && serializedModel.stepIdx > 0 && model.stepIdx < serializedModel.stepIdx + 20) {
```
pjanik: Also, we could leave a comment there that this new condition is important for authoring case, so it doesn't get removed in the future.

js/components/authoring.js
```diff
@@ -133,10 +165,14 @@ export default class Authoring extends PureComponent {
     const doneClass = done ? 'done' : ''
     const activeClass = idx === step ? 'active' : ''
     return (
-      <span className={`label ${activeClass} ${doneClass}`}>{ info }</span>
+      <span className={`label ${activeClass} ${doneClass}`} key={'info' + idx}>{ info }</span>
     )
   }
 
+  renderDivider (idx) {
+    return <div className='divider' key={idx} />
```
ekosmin: I realized my last PR left a duplicate key error because of the new children in an array. These key changes fix that

js/components/authoring.js
```diff
@@ -43,36 +47,38 @@ export default class Authoring extends PureComponent {
...
   handleBackButtonClick () {
     const { step } = this.state
+    const { restoreLabeledSnapshot } = this.props
     if (step === 2) {
       this.unloadModel()
     } else if (step === 3) {
+      restoreLabeledSnapshot(STEP_NAMES[2])
       this.setStep2()
+    } else if (step === 4) {
+      restoreLabeledSnapshot(STEP_NAMES[3])
```
ekosmin: I opted for restoring steps by number explicitly, rather than making a function that restores the "last" step. In the future, if multiple steps can be skipped, going to the "last" numerical step might be incorrect.

js/plates-model/model-worker.js
```diff
@@ -97,6 +98,14 @@ onmessage = function modelWorkerMsgHandler (event) {
   } else if (data.type === 'restoreInitialSnapshot') {
     self.m = model = Model.deserialize(initialSnapshot)
     snapshots.length = 0
+  } else if (data.type === 'takeLabeledSnapshot') {
```
ekosmin: I went back and forth on this, but I settled on making a new worker call for taking a labeled snapshot. I considered implementing a single call that would pick a storage object based on whether or not a label was passed...but the storage and restoring works so differently that I didn't want to overload things.

pjanik:
Changes look good, but there's something wrong with this PR I guess. It's open against `density-selector` branch instead of `master`, plus there're some conflicts, so it can't be merged. Maybe you can create a new branch, cherry-pick necessary commits, and open new PR?

ekosmin:
Good point! I did intentionally open it against `density-selector`, since it relies on those changes. Was that a reasonable thing to do?

I've rebased it off the updated branch so it should merge cleanly now.

PR #7, opened 2017-10-20
Adding density-setting as a configurable parameter

ekosmin:
Density-setting is now hidden unless the URL parameter 'setDensities' is passed. This means conditionally hiding the step in the bottom bar, and skipping it after step 3. I held off on introducing an elegant way to hide additional steps, but if we added a second configurable step it's probably worth looking into.

### DIFF
--- a/js/components/authoring.js
+++ b/js/components/authoring.js
@@ -6,6 +6,7 @@ import presets from '../presets'
 import ccLogo from '../../images/cc-logo.png'
 import ccLogoSmall from '../../images/cc-logo-small.png'
 import SortableDensities from './sortable-densities'
+import { getURLParam } from '../utils'
 
 import '../../css/authoring.less'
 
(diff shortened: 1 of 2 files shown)


js/components/authoring.js
```diff
@@ -180,9 +194,14 @@ export default class Authoring extends PureComponent {
           <div className='divider' />
           { this.renderStep(3) }
           { this.renderInfo(3, 'Assign forces to plates') }
-          <div className='divider' />
-          { this.renderStep(4) }
-          { this.renderInfo(4, 'Order plates by density') }
+          {
+            this.densityStepEnabled &&
```
pjanik: Note that there's a config object available:
https://github.com/concord-consortium/plate-tectonics-3d/blob/master/js/config.js
It automatically handles URL params to overwrite default values. Obviously, URL param needs to have the same name as config property.

So you could just add `densityStepEnabled` to this config object and just import `config` here and type `config.densityStepEnabled &&`. No need to parse URL param by hand.

ekosmin: Much cleaner this way, thanks Piotr!

PR #10, opened 2017-10-24
Density selector

ekosmin:
Here is a first pass at the density selector! Basically, I added the ability to set densities in the model, and pass that information to the model proxy and meshes. The densities are set by a new component in authoring, which uses a color-coded sortable list.


js/plates-model/field.js
```diff
@@ -241,12 +241,13 @@ export default class Field extends FieldBase {
       return
     }
 
-    if (this.isOcean && this.density < field.density) {
+    // Higher density plates subduct
+    if (this.isOcean && this.density > field.density) {
```
ekosmin: This was one big change I made, which seems to have broken the serialization tests. Previously, it seems like lower density plates were subducting instead of higher density plates, which I believe is backwards. These changes, along with a small rendering change, should fix that

js/components/sortable-densities.js
```diff
@@ -0,0 +1,80 @@
...
+
+function hsvToBackground (hsvColor) {
+  let rgb = hsv(hsvColor.h, hsvColor.s, hsvColor.v).rgb()
+  return {backgroundColor: 'rgb(' + Math.floor(rgb.r) + ', ' + Math.floor(rgb.g) + ', ' + Math.floor(rgb.b) + ')'}
+}
+
+const SortableItem = SortableElement(({plateInfo}) =>
+  <li className='density-button-container' style={hsvToBackground(plateInfo.color)}>
+    <div className='shading-box'>
+      <div className='density-button' key={key++}> {plateInfo.label} </div>
```
pjanik: 1. I think `key` should be always applied to the top element. I've checked console and react is complaining about missing keys.
2. I'd say that assigning keys like this is against the purpose of the `key` property. It should be used to identify which elements has changed and can be updated, which need to be removed and which need to be added to DOM. Here, it's pretty similar to saying `key=Math.random()`. React will remove and re-add all the elements on each render pass. It's not a big here, as we have 5 buttons usually. But in general, this approach might cause issues e.g. with performance and animations. So, I believe it's good to keep it right. Key should identify an element in a unique way, so react knows what to do with its DOM. The obvious candidate here would be `plateInfo.id`.

js/components/authoring.js
```diff
@@ -44,14 +47,23 @@ export default class Authoring extends PureComponent {
...
       this.setState({step: 3})
     } else if (step === 3) {
+      setOption('interaction', 'none')
+      setOption('selectableInteractions', [])
+      setOption('colormap', 'plate')
+      this.setState({step: 4})
+    } else if (step === 4) {
+      const { setDensities } = this.props
+      setDensities(this.state.intermediateDensities)
+      setOption('colormap', 'topo')
```
pjanik: Not necessary, colormap is set a few lines below to config.colormap.

js/components/plates.js
```diff
@@ -362,7 +370,9 @@ export default class Plates extends PureComponent {
         </div>
         {
           authoring &&
-          <Authoring loadModel={this.loadModel} unloadModel={this.unloadModel} setOption={this.handleOptionChange} />
+          <Authoring loadModel={this.loadModel} unloadModel={this.unloadModel}
+            setDensities={this.setDensities} setOption={this.handleOptionChange}
+            plates={this.modelProxy.plates} />
```
pjanik: We shouldn't do that in React. Props should be based only on parent component's state and props only. If `modelProxy.plates` is changed, React won't know that it should trigger re-rendering. I know that in this particular case, probably it won't happen. Or re-rendering will be triggered by some other property update. But in general, such approach can lead to bugs that are pretty hard to detect.

Also, I think we have too much state related to plate densities in general:
1. This property
2. intermediateDensities in Authoring
3. internal state of SortableDensities component

I think it could be reduced to one state kept at the top level. What I'd do:
- add `plateDensities` to `Plates` (this component) React state.
- `platesDensity` state is updated when there's a new message from the model worker, so it's always kept in sync with the model.
- pass it down to Authoring and then down to SortableDensities.
- when sorting happens, SortableDenisties triggers a callback that updates `platesDensity` here.
- update of `plateDensities` here triggers postMessage to the model worker. Make sure that this message is sent only if there's any difference in densities, so we avoid an infinite cycle of messages between this component and model worker.

I know it's more typing and seems more complicated, but it feels to me that it's a correct approach without shortcuts. And if we follow this pattern for new things in the future, hopefully, it'll save some debugging time.

js/plates-model/field.js
pjanik: btw, presets need to be updated too (all the densities need to be inverted):
https://github.com/concord-consortium/plate-tectonics-3d/blob/master/js/presets.js
so they work as they used to.

pjanik: Re failing tests, I'd try to update densities in test case too. It shouldn't fail in general. But if it won't fail after densities update, it means that this problem is probably not related to this PR. Also, a test should fail when you invert densities in the master branch. 😉

js/components/sortable-densities.js
ekosmin: Done, and good point! I'll keep that in mind for the future.

js/plates-model/generate-plates.js
```diff
@@ -32,10 +32,10 @@ export default function generatePlates (imgData, initFunction) {
...
     plates[key].addField({ id: fieldId, age: MAX_AGE, type, elevation, crustThickness })
   })
-  Object.values(plates).forEach(plate => plate.updateInertiaTensor())
+  Object.keys(plates).map(key => plates[key]).forEach(plate => plate.updateInertiaTensor())
   // User-provided function that can modify default options of all the plates.
   if (initFunction) {
     initFunction(plates)
   }
-  return Object.values(plates)
+  return Object.keys(plates).map(key => plates[key])
```
ekosmin: Tests were failing for me locally when using Object.values, so I swapped it out here. Apparently Object.values doesn't work on IE either, so this may be preferred

test/model-serialization.test.js
```diff
@@ -16,8 +16,12 @@ function getImageData (src, callback) {
 }
 
 function initFunc (plates) {
-  // const bluePlate = plates[210] // 210 hue
+  const bluePlate = plates[210] // 210 hue
   const yellowPlate = plates[70] // 70 hue
+  const purplePlate = plates[300] // 300 hue
```
ekosmin: Updating the densities here did correct the broken test...even though I'm still not sure why the test broke in the first place

js/components/plates.js
```diff
@@ -65,7 +65,9 @@ export default class Plates extends PureComponent {
       renderEulerPoles: config.renderEulerPoles,
       renderBoundaries: config.renderBoundaries,
       renderLatLongLines: config.renderLatLongLines,
-      snapshotAvailable: false
+      snapshotAvailable: false,
+      plateDensities: {},
+      plateColors: {}
```
ekosmin: Ended up needing state for the plate colors to pass to the density selector, does this seem like the best way to do that?

pjanik: It seems fine to me. Otherwise, we could keep `plates` array which would be sorted by densities and would include more information (e.g. ids, colors, and whatever is necessary in the future). Either way is fine with me.

js/plates-model/generate-plates.js
pjanik: I guess you might be using pretty old Node.js version. And the code is transpiled by Babel, so my guess is that Object.values gets replaced with something else or there's polyfill. But that's fine anyway. 👍

js/plates-model/model.js
```diff
@@ -102,6 +102,13 @@ export default class Model {
     this.setPlatesProp('angularVelocity', map)
   }
 
+  setDensities (densities) {
+    this.forEachPlate(plate => {
+      plate.setDensity(densities[plate.id])
+    })
+    this.plates.sort(sortByDensityDesc)
```
pjanik: I think this is why cross section might be broken (and some other things in the model).
Now, when there's swapped order of densities (hight numbers go to the bottom), I guess we should also swap order of plates array and sort them using `sortByDensityAsc`. There's an assumption that the first plate in this array is the top-most one. It's stated in the constructor, but we might add more comments.

js/presets.js
```diff
@@ -83,11 +83,11 @@ export default {
...
       const greenPlate = plates[130] // 130 hue
-      bluePlate.density = 0
-      violetPlate.density = 1
+      bluePlate.density = 4
+      violetPlate.density = 3
       yellowPlate.density = 2
-      greenPlate.density = 3
-      pinkPlate.density = 4
+      greenPlate.density = 1
+      pinkPlate.density = 0
```
pjanik: Still many presets are broken. Note that you need to apply similar changes even to presets where densities were not specified, and the defaults densities equal to IDs were used. We can't use defaults anymore in those cases.

ekosmin:
Thanks Piotr! I've double-checked all the presets, and their behavior looks like it matches now. Even in cases where it shouldn't really matter (e.g. divergent plates test) I explicitly set the density anyway. Re-sorting the array did seem to fix all the problems :)

js/components/sortable-densities.js
```diff
@@ -10,7 +10,7 @@ function hsvToBackground (hsvColor) {
 }
 
 const SortableItem = SortableElement(({plateInfo}) =>
-  <li className='density-button-container' style={hsvToBackground(plateInfo.color)} key={plateInfo.id}>
+  <li className='density-button-container' style={hsvToBackground(plateInfo.color)}>
```
ekosmin: @pjanik, I *believe* this key is unnecessary. I don't see any React warnings in my console, but I remember that you might have, so let me know how this looks for you.

pjanik:
Thanks! There were some conflicts related to my islands tweaking, so I just went ahead and resolved them.

PR #12, opened 2017-10-27
Add reset camera buttons for the main view and the cross section view

pjanik:
This PR adds reset camera buttons for the main view and the cross-section view.

I was going back and forth and I wasn't sure if we should keep the state of the cross-section camera reset in the main component or in the CrossSection3D. I implemented both and decided the latter makes more sense. First, it's pretty UI-specific state and I don't think that it's crucial for the app to know it (even if we used Redux or MobX). Second, `CrossSection3D` already contains camera-related state in the canvas views. So, I think that self-contained component makes more sense.

Also, I think that probably it would be nice to refactor `Plates` component and separate out main 3D view component, so it's similar to the `CrossSection` component.

### DIFF
--- a/js/components/cross-section-3d.js
+++ b/js/components/cross-section-3d.js
@@ -1,26 +1,56 @@
 import React, { PureComponent } from 'react'
 import CrossSection3DView from '../plates-view/cross-section-3d'
+import SmallButton from './small-button'
+
+import '../../css/cross-section-3d.less'
 
 export default class CrossSection3D extends PureComponent {
+  constructor (props) {
+    super(props)
+    this.state = {
+      showCameraResetButton: false
+    }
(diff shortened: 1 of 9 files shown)


js/components/cross-section-3d.js
```diff
@@ -1,26 +1,56 @@
...
     this.view3dContainer.appendChild(this.view.domElement)
   }
 
-  componentDidUpdate (prevProps) {
-    const { data, swapped } = this.props
-    if (data !== prevProps.data || swapped !== prevProps.swapped) {
-      this.view.setCrossSectionData(data, swapped)
-    }
+  componentDidUpdate () {
+    this.view.setProps(this.viewProps)
```
ekosmin: This seems much cleaner, useful pattern to keep in mind!

PR #13, opened 2017-10-29
Cross-section and island collision tweaks

pjanik:
A few tweaks related to Amy's requests (see commit messages).
Cross-section offset issue was pretty difficult to find. I'm not a fan of cross-section code, as it's trying to do too much, but well, it's all based on feature requests. I've added comments, so hopefully, it's clear more or less.

### DIFF
--- a/js/plates-model/plate.js
+++ b/js/plates-model/plate.js
@@ -252,8 +252,8 @@ export default class Plate extends PlateBase {
       })
     }
 
-    // Old island field is turned into regular ocean.
-    island.isOcean = true
+    // Remove the old island field.
+    island.alive = false
   }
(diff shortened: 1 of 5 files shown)


js/plates-model/plate.js
```diff
@@ -252,8 +252,8 @@ export default class Plate extends PlateBase {
       })
     }
 
-    // Old island field is turned into regular ocean.
-    island.isOcean = true
+    // Remove the old island field.
+    island.alive = false
```
ekosmin: To be honest I don't *totally* understand the new plate-breaking behavior, but it's interesting that it only took such a small change!

PR #14, opened 2017-10-30
Make authoring steps fully configurable

pjanik:
PR simplifies authoring component and makes it fully customizable. No more hardcoded step numbers.

There's a new config param: `authoringSteps`. It can be overwritten using URL, e.g.:
- http://models-resources.concord.org/plate-tectonics-3d/branch/flexible-authoring/index.html?authoring=true&authoringSteps=[presets,continents]
- http://models-resources.concord.org/plate-tectonics-3d/branch/flexible-authoring/index.html?authoring=true&authoringSteps=[presets,forces,densities]

I used `[a,b,c]` syntax, as it seems pretty compact and easy to serialize and deserialize. I think regexp is strict enough to avoid problems in practice. Using typical `parameter[]=a&paramter[]=b&parameter[]=c` will be more annoying if you want to define it by hand (+ more annoying to deserialize).

There'll be more changes to the authoring component soon, as I'm trying to make it work together with `preset` parameter (so you can preload model instead of picking one). And it'll work exactly the same way with `modelId` when it's ready.

UPDATE:
I've added new commit to this PR which ensures that authoring mode can be used with preset parameter too. It means we can combine authoring and existing model. I've also made sure that it'll support `modelId` parameter when it's implemented (but we'll need to double check it). Demo:
http://models-resources.concord.org/plate-tectonics-3d/branch/flexible-authoring/index.html?authoring=true&preset=subduction&authoringSteps=[continents,densities]

### DIFF
--- a/js/plates-model/model-worker.js
+++ b/js/plates-model/model-worker.js
@@ -104,7 +104,6 @@ onmessage = function modelWorkerMsgHandler (event) {
     let storedModel = labeledSnapshots[data.label]
     if (storedModel) {
       self.m = model = Model.deserialize(storedModel)
-      delete labeledSnapshots[data.label]
     }
   }
   forceRecalcOutput = true
--- a/css/authoring.less
+++ b/css/authoring.less
@@ -43,6 +43,12 @@
       margin: 0 10px;
     }
 
+    .step {
+      display: flex;
+      flex: 1 1 auto;
+      align-items: center;
+    }
+
     .divider {
       display: inline-block;
       border-top: 1px solid black;
@@ -88,7 +94,6 @@
       }
     }
 
-
     .cc-logo-large, .cc-logo-small {
       height: 46px;
       margin: 0 5px 0 10px;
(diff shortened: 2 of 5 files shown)


js/plates-model/model-worker.js
```diff
@@ -104,7 +104,6 @@ onmessage = function modelWorkerMsgHandler (event) {
     let storedModel = labeledSnapshots[data.label]
     if (storedModel) {
       self.m = model = Model.deserialize(storedModel)
-      delete labeledSnapshots[data.label]
```
pjanik: It doesn't seem necessary and it was causing bugs in authoring mode (when I wanted to go back and forth a few times).

pjanik:
@ekosmin, before you merge it, please take a look at https://github.com/concord-consortium/plate-tectonics-3d/pull/16. I'm leaving this PR so it's easier to read diff, but it'd be nice to merge #16 instead.

ekosmin:
Closing this in favor of #16, which includes a useful naming change

PR #15, opened 2017-10-31
Fixes slowness when side menu is opened

ekosmin:
Previously, every update from the worker updated density and color state stored in the Plates component. This caused constant re-renders to occur while the side menu was opened. Now, the state of Plates is only updated by the model once, upon initialization.


js/components/plates.js
```diff
@@ -109,6 +109,7 @@ export default class Plates extends PureComponent {
     this.restoreSnapshot = this.restoreSnapshot.bind(this)
     this.restoreInitialSnapshot = this.restoreInitialSnapshot.bind(this)
     this.handleResize = this.handleResize.bind(this)
+    this.isStateObjectEmpty = this.isStateObjectEmpty.bind(this)
```
pjanik: Not necessary unless you pass this function via reference somewhere (e.g. as a property or event handler).

js/components/plates.js
```diff
@@ -266,11 +267,16 @@ export default class Plates extends PureComponent {
     } else if (data.stepIdx === 0 && snapshotAvailable) {
       this.setState({snapshotAvailable: false})
     }
-    this.setDensities(this.convertPlatesToDensities(data.plates))
-    this.setState({plateColors: data.plates.reduce(function (plateIdToColor, plate) {
-      plateIdToColor[plate.id] = plate.baseColor
-      return plateIdToColor
-    }, {})})
+    // Plate densities and colors are initialized by the model, but never changed by it
+    if (this.isStateObjectEmpty('plateDensities')) {
```
pjanik: This doesn't sound right to me. It's an assumption that might bite us in the future. And I'm pretty sure it will fail when you use authoring mode, pick one preset, go next, then go back, and pick another preset. Densities and colors won't be updated.

pjanik: I guess what we should do here is to check equality instead.

pjanik:
I've been looking at it more. The previous densities and plate color update was definitely wrong.
It was causing that React component was updated on every single message from the model worker. I've added a log message to `componentDidUpdate` to confirm that. When you comment out densities and plate colors update, `componentDidUpdate` isn't called when new data from worker arrives.

In other words, we're triggering React re-render cycle on every message from the model worker. It's still not strictly related side menu, but well, it's a bit of UI and since it was re-rendered all the time, it could cause slowness on your machine (and perhaps some others).

If you use equality check, I think that it should solve the problem.

ekosmin:
Thanks for thinking through this @pjanik! Good point about the problematic assumptions. I now do the updates based on an equality check, while still preventing model updating loops. I did add Underscore to the project to get deep equality checking, let me know if you have another preference.

I also snuck in a second commit that fixes an unrelated sorting bug I found in my density component.

kswenson:
`lodash` is used more often than `underscore` in other CC projects (and in the broader JS community), so unless there's a particular reason to prefer `underscore` for this, `lodash` might be a better choice.

pjanik:
I'd agree with Kirk. Or actually, if it's just for comparison, I'd just use some tiny deep equal npm package or write it myself, as it's a few lines. It's not even a deep equal (just one level).

ekosmin:
Thanks @kswenson and @pjanik! FWIW the equality check is two levels deep since `plateColor` is an HSV object...I did originally write my own equality check, but figured a library would be simpler than writing something more flexible. I could try a deep equal library like [this one](https://www.npmjs.com/package/deep-equal), but I feel like we might have uses for `lodash` in the future? I don't have a good feel for how it affects performance to pull it in.

kswenson:
You can also use techniques like those discussed in https://stackoverflow.com/a/35251059 to just import the function(s) you need from lodash.

ekosmin:
Oh great! Based on the comments at that link, Webpack 2 uses tree shaking (a new term for me) so that `import { has } from 'lodash'` should only import the necessary file. I assume it's alright for us to rely on that?

ekosmin:
In any case, now using `lodash` instead of `underscore`!

pjanik:
@kswenson, thanks, that's really nice trick. I was reluctant to add the whole lib just for one function, but that completely solves the problem (and lodash funcs are nice).

@ekosmin, 
> Based on the comments at that link, Webpack 2 uses tree shaking (a new term for me) so that import { has } from 'lodash' should only import the necessary file. I assume it's alright for us to rely on that?

Not really, even though we use Webpack 3. I've checked that and `import { isEqual } from 'lodash'` was adding 0.6MB to the built file. `import isEqual from 'lodash/isequal'` added pretty much nothing. Maybe I'm missing some optimization flags, but I'd play safe and use `lodash/<filename>` syntax in the future. I've pushed a commit to fix that.

js/components/sortable-densities.js
```diff
@@ -35,7 +35,7 @@ export default class SortableDensites extends Component {
 
     // Aftering a reload, plate IDs continue to increment
     // Subtracting the smallest ID ensures visible plate numbering starts at 1
-    let minId = Object.keys(plateDensities).sort()[0] - 1
+    let minId = Object.keys(plateDensities).reduce(function (a, b){ return parseInt(a) < parseInt(b) ? a : b }) - 1
```
pjanik: @ekosmin, `sort` also accepts comparison function, but that is fine too (fancy). ;) 
You could even make it shorter using es6 syntax:
`reduce((a, b) => parseInt(a) < parseInt(b) ? a : b)`
😉

ekosmin: Thanks for running the check on the different import styles, I'll keep this in mind! And true, would have been a good use of es6

PR #16, opened 2017-10-31
Flexible authoring + renaming: authoring -> planetWizard

pjanik:
@ekosmin, this is this https://github.com/concord-consortium/plate-tectonics-3d/pull/14 + renaming: authoring => planetWizard.

Why renaming? All the other projects use ?authoring param for URL authoring (what we used to call classroom authoring page), including seismic-explorer which will be used together with this project. So, I think it makes sense to keep things consistent and stick to `authoring` param for URL authoring.

If you can think of some better name than `planetWizard` or if it sounds awkward, let me know. 😉 

It makes sense for review https://github.com/concord-consortium/plate-tectonics-3d/pull/14, as you should see changes in a cleaner way. And if you're fine with renaming, you can just merge this PR instead.

I think it's good to do it ASAP, so we have fewer conflicts in the upcoming PRs.

### DIFF
--- a/js/components/authoring.js
+++ b/js/components/authoring.js
@@ -17,7 +17,7 @@ const AVAILABLE_PRESETS = [
   { name: 'plates5Uneven', label: '5 plates', info: 'uneven distribution' }
 ]
 
-const STEPS = {
+const STEPS_DATA = {
   presets: {
     info: 'Select layout of the planet'
   },
@@ -32,12 +32,19 @@ const STEPS = {
   }
 }
 
+// When there's preset or modelId provided, make sure that preset selection step isn't used.
+// It's for authors convenience, so it's not necessary to modify default list of authoring steps
+// when preloaded model is used in authoring.
+const STEPS = config.preset || config.modelId
+  ? config.authoringSteps.filter(stepName => stepName !== 'presets') : config.authoringSteps
+
+console.log(STEPS)
+
 export default class Authoring extends PureComponent {
   constructor (props) {
     super(props)
     this.state = {
-      step: 0,
-      modelLoading: false
+      step: 0
     }
     this.handleNextButtonClick = this.handleNextButtonClick.bind(this)
     this.handleBackButtonClick = this.handleBackButtonClick.bind(this)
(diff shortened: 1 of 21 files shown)


js/components/authoring.js
```diff
@@ -17,104 +17,129 @@ const AVAILABLE_PRESETS = [
...
   }
 
   loadModel (presetInfo) {
     const { loadModel } = this.props
-    loadModel(presetInfo.name)
-    this.setStep2()
+    this.setState({ modelLoading: true })
+    loadModel(presetInfo.name, () => {
+      // Important! Go to the next step after model load is complete.
+      // Otherwise, we might try to take model snapshot too early (when it's not loaded yet).
```
ekosmin: Nice fix, this definitely wasn't addressed before

js/components/authoring.js
```diff
@@ -17,104 +17,129 @@ const AVAILABLE_PRESETS = [
...
-  setStep3 () {
+  setForcesStep () {
     const { setOption } = this.props
     setOption('interaction', 'force')
     setOption('selectableInteractions', [ 'force', 'none' ])
     setOption('colormap', 'topo')
-    this.setState({step: 3})
+  }
+
+  setDensitiesStep () {
```
ekosmin: Again, we could put these in the `STEP` object if we wanted, and further clean up `componentDidUpdate`. that's what I would probably do if this was Java, but I'm not sure what best practices are for js

js/components/plates.js
```diff
@@ -179,40 +176,41 @@ export default class Plates extends PureComponent {
...
+    const { modelState } = this.state
+    // Most of the messages require model to exist. If it doesn't, queue messages and send them when it's ready.
+    if (modelState === 'loaded' || data.type === 'load' || data.type === 'unload') {
+      this.modelWorker.postMessage(data)
+    } else {
+      this.modelMessagesQueue.push(data)
+    }
+  }
+
+  postQueuedModelMessages () {
```
ekosmin: This queueing looks good...it will be especially nice not to have to wait for loading to complete when we're just skipping through steps

js/components/authoring.js
```diff
@@ -32,12 +32,19 @@ const STEPS = {
   }
 }
 
+// When there's preset or modelId provided, make sure that preset selection step isn't used.
+// It's for authors convenience, so it's not necessary to modify default list of authoring steps
+// when preloaded model is used in authoring.
+const STEPS = config.preset || config.modelId
+  ? config.authoringSteps.filter(stepName => stepName !== 'presets') : config.authoringSteps
+
+console.log(STEPS)
```
ekosmin: This might have snuck in the commit

js/components/index-page.js
```diff
@@ -8,8 +8,8 @@ export default class IndexPage extends PureComponent {
     return (
       <div className='index-page'>
         <h1>3D Plate Tectonics models</h1>
-        <h2>Use <a href={`${window.location.pathname}?authoring=true`} target='_blank' rel='noopener noreferrer'>
-          authoring mode
+        <h2>Use <a href={`${window.location.pathname}?planetWizard=true`} target='_blank' rel='noopener noreferrer'>
```
ekosmin: I think "Use *the* planet wizard" is a little cleaner

js/components/authoring.js
```diff
@@ -17,104 +17,129 @@ const AVAILABLE_PRESETS = [
...
-  get buttonDisabled () {
+  get nextButtonLabel () {
     const { step } = this.state
-    return step === 1
+    return step === config.authoringSteps.length - 1 ? 'finish' : 'next'
   }
 
-  handleNextButtonClick () {
+  get navigationDisabled () {
+    return this.currentStep === 'presets'
```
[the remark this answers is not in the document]
pjanik: Right, there might be more steps like this. Done. 👍

js/components/authoring.js
pjanik: Implementation has changed in the following commits and now it's addressed using a message queue. Risk of reviewing commit by commit. 😉

js/components/authoring.js
pjanik: Right, we could do that. It would be more "meta". I've looked at it, but there are still functions that have a bunch of "setOption" calls and can't be moved to STEP_DATA, e.g. initial component setup or setup after the wizard is finished. Also, some steps do more work, e.g. presets step unloads the model. So, I've finally left it as it is. But it's an option for sure if number of steps grows or we want to do more refactoring.

js/components/index-page.js
pjanik: Thanks, fixed.

pjanik:
Thanks, @ekosmin. I've addressed some comments and also added back button fix.

PR #18, opened 2017-11-03
Making drag interface on density sorter more obvious

ekosmin:
@pjanik, re: the discussion [here](https://www.pivotaltracker.com/story/show/152289874), these are the changes I made to improve the UI of the density picker. Like Amy pointed out, this doesn't do much in terms of scaling on different devices, but I'm not sure the best way to handle that. In Geniverse, we just applied a scale factor on top of the whole page, but I realize you've made GEODE more responsive, so I'm curious what your thoughts are.

Specifically, the changes are: adding a hamburger menu and changing cursor to indicate drag-ability. Also adding helper text.

### DIFF
--- a/css/planet-wizard.less
+++ b/css/planet-wizard.less
@@ -178,5 +178,13 @@
     align-items: start;
     justify-content: center;
     margin-left: 13%;
+
+    .helper-text {
+      width: 200px;
+      text-align: center;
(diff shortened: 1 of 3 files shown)


ekosmin:
Much easier with the FontIcon! Fixed up the code and merged, thanks @pjanik

PR #19, opened 2017-11-07
Authoring

pjanik:
This PR adds a new page available at `?authoring`. It lets authors customize URL for an activity.
Sidebar panel is now configurable too.

Also, I've started using css-modules, so I guess we could slowly switch to that and maybe refactor other components while working on them. Actualy, sometimes it's the only way to theme react-toolbox components, so that's why I had to use it. But I think it's a nice pattern in general.

### DIFF
--- a/css-modules/authoring.less
+++ b/css-modules/authoring.less
@@ -1,0 +1,88 @@
+.authoring {
+  font-size: 16px;
+  padding: 20px;
+
+  hr {
+    width: 300px;
+    margin: 10px 0;
+    border: 1px solid #eee;
+  }
+
+  h2 {
+    margin-bottom: 10px;
+  }
+
+  h3 {
+    margin-top: 20px;
+  }
+
+  iframe {
+    border: 1px solid #aaa;
+  }
+
+  .inlineInput {
+    display: inline-block;
+    margin-left: 10px;
+    margin-right: 10px;
+  }
+
+  .section {
+    margin-left: 10px;
+  }
+
+  .finalUrl {
+    border: 1px solid #999;
+    padding: 10px;
+    border-radius: 3px;
+  }
+
+  .autocompleteLabel {
+    font-size: 14px;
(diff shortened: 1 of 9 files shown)


css-modules/authoring.less
```diff
@@ -0,0 +1,88 @@
...
+    color: #666;
+    top: 24px;
+  }
+
+  .filled ~ .label {
+    color: #555;
+  }
+}
+
+// checkbox theme
```
ekosmin: I read up on CSS modules - having scoped CSS does seem really nice, I'll try it in my next commit. I guess if the toolbox components are written with css modules, the class names all end up gibberish, so you need to use css modules to style them?

pjanik: Yes, exactly. Sometimes it's enough to assign className to the top component and then try to use element matches, but quite often it's annoying and I reached the point that I couldn't do that. With CSS modules, you just specify class names that are described in react-toolbox docs and you can theme all the elements.

It makes sense and it won't hurt our code either. No need to care about unique class names.

pjanik:
@ekosmin, good catch, thanks! I was testing with localhost:8080/ but ofc it didn't work with index.html. Fixed now.

I had to force push again and merge master into authoring branch again. Not sure how you merged changes, but it looked strange and you force pushed your changes, so I got in a state where I couldn't push my changes at all. Not sure why, but it wasn't worth more investigation. Nice way to do it is to merge master into topic branch, so they no force pushing is necessary. Github website has pretty nice UI for it actually, so there's even no need to do it locally.

ekosmin:
Sorry about the forced changes, I've never used the Github merge tools before, but I'll give them a try next time! I think this looks good now - I had to change the default value of `modelId` since otherwise there was a NPE from a missing constructor. Do we want to be able to accept null defaults in config? Either way, I'm going to go ahead and merge this.

PR #20, opened 2017-11-07
Add step forward button

pjanik:
Pretty simple one.

I started implementing targetStepIdx as I did in 2D model:
https://github.com/concord-consortium/plate-tectonics/blob/master/js/plates-model/model.worker.js#L34-L37

A single step in 2D model was taking way longer, so it was useful there. But then I checked that simple approach works fine here, as a step calculation is faster.


ekosmin:
Simple enough, LGTM!

PR #21, opened 2017-11-13
Continent "buffer", performance optimization, field / crust type cleanup

pjanik:
I think it makes sense to review commits separately. This PR adds a few different things:

1. Continent "buffer" area that prevents / limits subduction of continents.
2. The width of the oceanic ridge is increased.
3. Improves performance of island detection. It's not necessary to calculate it in every step. However, continent buffers need to be calculated in every step as we keep adding a new oceanic crust. But I spent a while on optimization and now it's pretty fast. If you compare master branch with this one (using &benchmark param), this one should be still faster even though these new calculations have been added.
4. I've tried to clean up field / crust type. There's a new property that specifies field type and separate getters for crust type.


js/plates-model/field.js
```diff
@@ -251,6 +259,10 @@ export default class Field extends FieldBase {
         this.volcanicAct = new VolcanicActivity(this)
       }
       this.volcanicAct.setCollision(field)
+      if (this.isOcean && field.isContinentBuffer) {
```
ekosmin: Is this check necessary in addition to the one above?

js/plates-model/plate.js
```diff
@@ -185,6 +185,42 @@ export default class Plate extends PlateBase {
...
+            queue.push(adjField)
+          }
+        })
+      }
+    })
+
+    while (queue.length > 0) {
+      const field = queue.shift()
+      field.isContinentBuffer = true
+      const newDist = getDist(field) + grid.fieldDiameterInKm
```
ekosmin: Out of curiosity, what is the reason for using kilometers to measure distance instead of hexes?

js/plates-model/field.js
```diff
@@ -263,14 +291,14 @@ export default class Field extends FieldBase {
         // Special case when the continent is "trying" to subduct under the ocean. Apply drag force to stop both plates.
         this.draggingPlate = field.plate
       }
-    } else if (this.isContinent && field.isContinent && !field.island) {
+    } else if (this.isContinent && field.isContinent) {
```
ekosmin: It seems like the effect of this refactor is that old checks for `isContinent` should become checks for `continentalCrust` and `isContinent && !isIsland` should become `isContinent`. Following that logic, I would expect this check to read `else if (this.continentalCrust && field.isContinent)`

js/plates-model/field.js
pjanik: Which one above? The similar one in the first `if` statement? 
If so, then yes, it's necessary as one would apply force to the top plate, and the second one to the bottom plate.

ekosmin: OK I see, they both drag each other. Is this preferable to adding `field.draggingPlate = this.plate` in the first `if` statement?

js/plates-model/field.js
pjanik: Good catch! Although, I think that `&& !field.island` was unnecessary. Island-continent collision is handled above and this specific collision type is only applied to "real" continents (orogeny / mountains forming).

btw, another candidate for refactoring is this collision logic. I guess it would be nice to get rid of `else` key word and make sure that all the conditions are self-sufficient and don't depend on the others. But then it'd be nice to check if it affects performance in a visible way or not.

js/plates-model/field.js
pjanik: It's a bit more complicated. Note that collideWith handles only side of collision. As currently field A can collide with field B, but B might not collide with field A (due to optimized collision detection). So, yes, we could do what you suggested, but it'll be just a bit different behaviour. Not better, not worse.
 
But it's very reasonable question and it made me start working on refactoring of this part. I got to some way cleaner implementation where collision is handled for the pair (not single field like now). But it works a bit different and I'm currently testing if it's a problem or not. I'll update this PR when it's ready.

js/plates-model/plate.js
pjanik: Note that hexes can have different dimensions. It's configurable using `divisions` param. So, the grid can be denser. And we don't want more precise model to behave differently. 

In general, I think it's always better to use some kind of world units in modeling, whenever it's possible. Or come up with some model unit system. Easier to read, understand, and it's more independent of the implementation. I can say in config that this buffer area is 750km wide, no matter what grid (or even model) we use.

js/plates-model/field.js
pjanik: Done in this commit: 
https://github.com/concord-consortium/plate-tectonics-3d/commit/54ac5f9a13aabbe23f6914ae8d51db0e652c88e5

I hope it looks cleaner now and also it's faster as we limit the number of checks. It affects model behavior a bit, but it doesn't seem that it's better or worse. It's just a bit different.

pjanik:
@ekosmin, I pushed one more refactoring. Thanks for good review, I think I replied to all the comments now.

js/plates-model/fields-collision.js
```diff
@@ -0,0 +1,60 @@
...
+  if (bottomField.isOcean) {
+    subduction(bottomField, topField)
+    if (bottomField.isContinentBuffer && topField.isOcean) {
+      // Special case when the continent is "trying" to subduct under the ocean. Apply drag force to stop both plates.
+      applyDragForces(bottomField, topField)
+    }
+  } else if (bottomField.isContinent) {
+    if (topField.isContinent) {
+      orogeny(bottomField, topField)
+    } else { // topField.isOcean || topField.isIsland
```
ekosmin: It seems like this `else` case could almost be merged with the `bottomField.isContinentBuffer && topField.isOcean` case above. If the condition was instead `(bottomField.isContinentBuffer || bottomField.isContinent) && (topField.isOcean || topField.isIsland)`, the comment in both blocks wouldn't need to be duplicated. The only change is that this would create drag forces when a continent buffer collides with an island, which seems like it could be a desirable change?

js/plates-model/fields-collision.js
```diff
@@ -0,0 +1,60 @@
...
+
+  if (bottomField.isOcean) {
+    subduction(bottomField, topField)
+    if (bottomField.isContinentBuffer && topField.isOcean) {
+      // Special case when the continent is "trying" to subduct under the ocean. Apply drag force to stop both plates.
+      applyDragForces(bottomField, topField)
+    }
+  } else if (bottomField.isContinent) {
+    if (topField.isContinent) {
+      orogeny(bottomField, topField)
```
ekosmin: So to clarify, this is the orogeny change - continents trying to subduct under islands now creates drag forces, instead of orogenies.

js/plates-model/plate.js
ekosmin: That makes a lot of sense, thanks for the explanation!

js/plates-model/fields-collision.js
pjanik: You're right. It wasn't planned change, but this case is only theoretical at the moment. We never have islands at the boundary, as they're created in some distance from it. We don't have such presets either. And the drag forces are applied in orogeny too. The only difference is whether some small mountains will be formed or not. I don't really have an opinion on that. We can change it to `if (topField.continentalCrust) ...` if it's necessary later or it feels better to you.

js/plates-model/model.js
```diff
@@ -167,17 +168,21 @@ export default class Model {
...
+    // so consistent as subduction should.
+    const platesCount = this.plates.length
+    for (let i = platesCount - 1; i >= 0; i -= 1) {
+      const bottomPlate = this.plates[i]
+      bottomPlate.forEachField(bottomField => {
+        for (let j = i - 1; j >= 0; j -= 1) {
+          const topPlate = this.plates[j]
+          const topField = topPlate.fieldAtAbsolutePos(bottomField.absolutePos)
+          if (topField) {
+            fieldCollision(bottomField, topField)
```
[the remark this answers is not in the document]
pjanik: Definitely, thanks!

js/plates-model/fields-collision.js
pjanik: > If the condition was instead (bottomField.isContinentBuffer || bottomField.isContinent) && (topField.isOcean || topField.isIsland)

Not sure where would we put it? Note that `if`s are now nested. The first level checks bottom field and it's very simple, no && and ||. Then we look at the top field. I find it a bit easier to reason about.

> The only change is that this would create drag forces when a continent buffer collides with an island, which seems like it could be a desirable change?

Yes, that makes sense. It'll be consistent with the logic related to orogeny (continent vs continent => orogeny, contient / cont. buffer vs non-continent => only drag force).

pjanik:
Thanks @ekosmin! I've added new commits to address some issues.

pjanik:
I think all the comments are addressed now, except one that is not clear to me, so I'll go ahead and merge it.

PR #22, opened 2017-11-19
Increase ocean depth to have more obvious relief at ocean boundaries

pjanik:
Updates some defaults, all the preset images, and colormaps.


pjanik:
@ekosmin, I'm merging it so Amy can play with it when we're away, but feel free to leave some comments when you're back and I'll be happy to address them. Thanks.

PR #23, opened 2017-11-19
Enforce relative motion between plates and break big plates into smaller chunks

pjanik:
Adds a new config option: `enforceRelativeMotion`. When it's true (default), the model will make sure that there's some relative motion between plates. Also, big plates (bigger than `config. minSizeRatioForDivision`) will be split into two smaller plates.


pjanik:
@ekosmin, I'm merging it so Amy can play with it when we're away, but feel free to leave some comments when you're back and I'll be happy to address them. Thanks.

PR #26, opened 2017-11-30
Seed random behaviour, add 'deterministic' option (+ add / update tests)

pjanik:
Add a new 'deterministic' option to config and ensures that model behaves in a deterministic way when it's set to true. It also works with model saving and loading, as the current seedrandom state is being serialized.

### DIFF
--- a/js/seedrandom.js
+++ b/js/seedrandom.js
@@ -1,0 +1,20 @@
+import seedrandom from 'seedrandom'
+
+const SEED = 'PlateTectonics3D'
+let rand = null
+
+export function initialize (deterministic) {
+  rand = seedrandom(SEED, { state: true, entropy: !deterministic })
+}
+
+export function initializeFromState (state) {
+  rand = seedrandom('', { state })
+}
+
+export function getState () {
+  return rand.state()
+}
+
+export function random () {
+  return rand()
+}
--- a/js/seedrandom.js
+++ b/js/seedrandom.js
@@ -4,10 +4,12 @@ const SEED = 'PlateTectonics3D'
 let rand = null
 
 export function initialize (deterministic) {
+  // state: true enables state saving, entropy controls whether random generator is deterministic or not.
   rand = seedrandom(SEED, { state: true, entropy: !deterministic })
 }
 
 export function initializeFromState (state) {
+  // When state is provided, the first argument, seed, is ignored.
   rand = seedrandom('', { state })
 }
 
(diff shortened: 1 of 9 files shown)


js/seedrandom.js
```diff
@@ -0,0 +1,20 @@
+import seedrandom from 'seedrandom'
+
+const SEED = 'PlateTectonics3D'
+let rand = null
+
+export function initialize (deterministic) {
+  rand = seedrandom(SEED, { state: true, entropy: !deterministic })
```
ekosmin: I had to look through the `seedrandom` readme to understand these constructors - it might be worth adding a couple of comments to clarify the parameters

test/model-serialization.test.js
```diff
@@ -113,16 +113,20 @@ test('model serialization', () => {
...
-    model2.step(TIMESTEP)
   }
-  compareModels(model2, model1)
+  // We can't use just one loop as random seed is set when model is created or deserialized. Creating model at this
+  // point ensures that both use the same seedrandom state.
+  const model3 = Model.deserialize(model1Data)
+  for (let i = 0; i < 200; i++) {
+    model3.step(TIMESTEP)
+  }
+  compareModels(model3, model1)
```
ekosmin: Should this be pulled out into a separate unit test, to check the randomness specifically?

test/seedrandom.test.js
```diff
@@ -0,0 +1,28 @@
...
+
+test('non-deterministic random', () => {
+  seedrandom.initialize(false)
+  const result1 = [seedrandom.random(), seedrandom.random(), seedrandom.random()]
+  seedrandom.initialize(false)
+  const result2 = [seedrandom.random(), seedrandom.random(), seedrandom.random()]
+  expect(result1).not.toEqual(result2)
+})
+
+test('save and restore random', () => {
```
ekosmin: Nice tests!

js/seedrandom.js
pjanik: Thanks, I'll do it.

test/model-serialization.test.js
pjanik: This test both checks randomness, but also it checks whether models have been serialized and deserialized correctly. If some property is missing, it's very likely the models will diverge after some time. So, the main purpose of this test is to check serialization. I thought about dividing it, but it's difficult to do that without making tests slower (and they're already pretty slow).

PR #27, opened 2017-12-01
Leaving a border of lower elevation around drawn continents

ekosmin:
Relatively simple change - I factored out the logic for drawing a tile, and draw a slightly lower tile on the first tile which is too far to draw an actual continent. This means that the drawn area is now slightly larger than before. I could have lowered the max `distance` from 7 to counteract this, but the behavior seems fine so I did not. 

The only complication was making sure that the max continent to plate ratio was maintained. Right now, you can overwrite the continental shelf once you hit the ratio, but I think that's mostly an edge case

### DIFF
--- a/js/plates-model/field.js
+++ b/js/plates-model/field.js
@@ -241,6 +241,16 @@ export default class Field extends FieldBase {
     }
   }
 
+  anyNeighbour (condition) {
+    for (let adjId of this.adjacentFields) {
+      const field = this.plate.fields.get(adjId)
+      if (field && condition(field)) {
+        return true
+      }
+    }
+    return false
+  }
+
   // One of the neighbouring fields, pointed by linear velocity vector.
   neighbourAlongVector (direction) {
     const posOfNeighbour = this.absolutePos.clone().add(direction.clone().setLength(grid.fieldDiameter))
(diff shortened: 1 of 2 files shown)


pjanik:
Looks good, thanks. I've found one small issue - it doesn't smooth out the edges when we reach the max continent area for a plate. I've been trying to fix it, but it's a bit tricky. Anyway, I'll update this PR when I come up with the right solution and merge it.

ekosmin:
@pjanik Yes it's definitely an issue - that's the complication I was trying to describe in my first message (which I did edit after writing, so the PR email probably didn't include it). Perhaps the fix is to disable drawing entirely once the max area is reached? Basically changing the first line of `drawField` to something like this:
```
if (type === 'continent' && (continentSize + 1) / plateSize > MAX_CONTINENTAL_CRUST_RATIO) {
    return
}
```
That would solve this problem, even though it would mean users couldn't update any continent heights until something was erased.

pjanik:
> That would solve this problem, even though it would mean users couldn't update any continent heights until something was erased.

I think that's fine. Thanks, that helped to fix one of the issues.
I've refactored drawing quite a bit and also made sure it works with erasing.

PR #28, opened 2017-12-01
Deforming top plate in subducting zones

ekosmin:
Here's the change for deforming the top plate - the modifier is a bit arbitrary, but a simple change otherwise


pjanik:
Looks good, thanks!

PR #30, opened 2017-12-28
Refactor codebase to use Mobx

pjanik:
This is probably too big to review, but opening this PR for the future reference.
It's an example of the transition from raw React code to MobX-based.


kswenson:
I won't claim to have performed a thorough review, but I did read through the changes/additions and they looked good to me. I think your observations about strengths/weaknesses of MobX and its performance characteristics -- both here and your Slack post on the topic -- will be useful to the rest of us moving forward. Nice work.

pjanik:
Thanks, @kswenson! :)  This PR is mostly for future reference. I didn't expect anyone to perform a thorough review, but I'm happy you took a look!

PR #31, opened 2018-05-29
Add splash screen

pjanik:
I created a separate webpack entry point and output script, so we can add splash screen before the main `app.js` finishes loading and being evaluated. It actually takes a long time, as the file is large and simulation grid is calculated on load. Using separate script, I could show splash screen before that finishes.


js/components/splash-screen.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+    this.onWindowResize = this.onWindowResize.bind(this)
+  }
+
+  componentDidMount () {
+    setTimeout(() => {
+      this.setState({ fadeOut: true })
+    }, HIDE_AFTER - TRANSITION_DURATION)
+    setTimeout(() => {
+      this.setState({ show: false })
```
ekosmin: Small, but I wonder if it's worth removing the event listener once the splash screen hides? Seems like it never unmounts so it just sticks around.

js/components/splash-screen.js
```diff
@@ -0,0 +1,60 @@
...
+    this.setState({ scale: Math.min(1, height / MIN_HEIGHT, width / MIN_WIDTH) })
+  }
+
+  render () {
+    const { show, fadeOut, scale } = this.state
+    if (!show) {
+      return null
+    }
+    return (
+      <div className={`${css.splashScreen} ${fadeOut ? css.fadeOut : ''}`}>
```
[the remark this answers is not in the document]
pjanik: Right, that's CSS modules. 😉 
I think it makes sense, less chance for conflicts, enforces better patterns. I introduced that in PortalPages too where conflicts are even more possible.

js/components/splash-screen.js
pjanik: Good catch, thanks.

I actually removed this listener at all. I was using it when I was playing with splash screen displayed all the time. But in reality, it's only on page load for a second or two. So, we can read window size on mount and just use it. I don't think anyone would be able to resize window while it's visible. 😉

pjanik:
> The only strange behavior I noticed is that the splash screen doesn't seem to load properly/at all if I soft refresh the page. Is that expected behavior?

Not really. Hopefully, I fixed the real problem now - `Grid` class initialization. It used to happen while the script was evaluated, which was causing long freeze and forced me to use a separate script. I changed that to a proper singleton that provides `getGrid` function. And this function is now used only when necessary in class methods. There's no freeze on load, index page loads faster, and I can use just one script to display the splash screen. Does it work better now?

pjanik:
@ekosmin, have you had a chance to check if the new version fixed soft reload issue for you?

ekosmin:
@pjanik Yes it looks like the soft reload behavior is consistent now! Pretty minor, but it looks like the screen flashes an unscaled version of the logo quickly before resizing.

pjanik:
> Pretty minor, but it looks like the screen flashes an unscaled version of the logo quickly before resizing.

@ekosmin, you're right, it was bothering me as well. It was caused by webpack CSS processing - CSS is inlined in JS, but then it's added as a separate link to blob to HTML page. So, CSS is always loaded after JS as this blob needs to be "downloaded" again. I made splash screen styles inline to eliminate this delay. It should look good now.

PR #33, opened 2018-06-05
Plate markers

ekosmin:
@pjanik The plate markers might keep changing, but I was hoping you could take a look at the code I have so far! I think the commits ended up pretty clean and straightforward.

http://tectonic-explorer.concord.org/branch/plate-markers/index.html

https://www.pivotaltracker.com/story/show/157632753


js/plates-view/plate-label.js
```diff
@@ -0,0 +1,62 @@
+import * as THREE from 'three'
+import { hueAndElevationToRgb } from '../colormaps'
+import config from '../config'
+
+const RADIUS = 1.025
+
+function pointTexture (label, color) {
```
ekosmin: The biggest change I was considering is refactoring this label to avoid overlap with the cross section points

pjanik: Blind guess, but if you play with the radius, it might be an easy tweak. Otherwise, we'd need separate rendering layer which is quite painful. I don't think this project is using that yet.

pjanik:
It looks very good. Maybe we could try to reuse the existing code a bit more (point generation)? But it doesn't seem very necessary.

Also, I'd check how that affects performance, you can use "benchmark" url param. Center is updated every step even if plate labels are disabled. I guess we could calculate it only when necessary in `model-output.js` (it has some conditionals based on rendering props). Although if this view is enabled by default and performance impact isn't huge, your solution is cleaner.

Also, not sure if serialization is supported (but I might have missed it).

js/plates-view/plate-label.js
ekosmin: Oh right sorry, I meant trying to refactor the duplicated code :)

ekosmin:
Thanks @pjanik! I checked the benchmark as you suggested, and it looks like these changes did slow things down 5-10 steps per second. As I [mentioned on PT](https://www.pivotaltracker.com/story/show/157632753/comments/190370342), I changed the algorithm for setting the point locations so they now only update if they fall off the plate. Since I'm not iterating over the fields every step now, I don't notice any dip in performance.

Also, I think serialization is working as expected. I added `center` to the list of `serializableProps`, and when I load a saved model, the labels are in the same place (rather than recalculated to the center).

I also factored out `PointLabel` into a separate class, it may be useful in the future.

js/plates-model/plate.js
```diff
@@ -85,12 +85,38 @@ export default class Plate extends PlateBase {
...
+      if (centerField && !centerField.colliding) {
+        return
+      }
+    }
+
+    // If the current label position is no longer valid, move it back to the center
     let sum = new THREE.Vector3(0, 0, 0)
     this.fields.forEach(field => {
-      sum = sum.add(field.absolutePos)
+      if (!field.colliding) {
```
ekosmin: @pjanik I struggled a lot trying to figure out how to only include "visible" fields in the calculations. I tried using `field.subduction`, but I think that missed some fields that were still under other plates? I ended up using `field.colliding`, but it's not perfect, because there are some visible fields colliding with others and skipping them throws off the calculations.

js/plates-model/plate.js
```diff
@@ -79,6 +84,41 @@ export default class Plate extends PlateBase {
...
+    this.fields.forEach(field => {
+      // Ensure we label a visible field
+      if (!field.colliding) {
+        const dist = field.absolutePos.distanceTo(center)
+        if (dist < minDist) {
+          closestPoint = field.absolutePos
+          minDist = dist
+        }
+      }
+    })
```
pjanik: This is not a very efficient way to calculate the closest point. You can get it in O(1):
https://github.com/concord-consortium/tectonic-explorer/blob/master/js/plates-model/grid.js#L94-L101
(once you have field ID, you can have field and it's absolute position too)
You should try to read the code of the whole class (it's not too long), and see how it's used in various places. It's core of the model grid and many many calculations, so it's very helpful to understand how it works. Basically, this helper lets you find the nearest field to a given position in O(1). This is the main thing used in the collision detections.

I wouldn't care about "colliding" property either, but I'll talk about it more in another comment.

pjanik:
Changes look good, thanks!

> @pjanik I struggled a lot trying to figure out how to only include "visible" fields in the calculations. I tried using field.subduction, but I think that missed some fields that were still under other plates? I ended up using field.colliding, but it's not perfect, because there are some visible fields colliding with others and skipping them throws off the calculations.

I think subduction should work. If a field is not visible, it has to be subducting. There's some subtle overlapping during orogeny, but I don't think it's significant. `colliding` will skip many other fields, e.g. ones that are on top of other plates (where volcanic activity happens).

Also, I'd consider recalculating center more often, e.g. every 1-2 seconds (we aim at 30 model steps per second, so it should be around 50-100 model steps). We woudn't really need to check if center is still visible or not. I think some other calculations could also use some longer interval in the future (as now almost everything is based on physics engine timestep). The model provides this property as "center", so it would be cool if we could assume that it's around center usually. ;)

ekosmin:
Hi @pjanik! Let me know if you have a chance to check out my recent changes - mainly the [updated center-finding code](https://github.com/concord-consortium/tectonic-explorer/blob/f12c2df18aada6941410d54861dde8cc80f754a6/js/plates-model/plate.js#L87) and [lowered recalculation rate](https://github.com/concord-consortium/tectonic-explorer/pull/33/commits/1c53f44e2d78cacaada9719eae453b03dda79f22). I'm fairly happy with this behavior on the complicated use-cases:

http://tectonic-explorer.concord.org/branch/plate-markers/index.html?preset=earth

http://tectonic-explorer.concord.org/branch/plate-markers/index.html?preset=islandCollisions

http://tectonic-explorer.concord.org/branch/plate-markers/index.html?preset=circles

js/plates-model/plate.js
```diff
@@ -79,6 +84,49 @@ export default class Plate extends PlateBase {
     return totalTorque.applyMatrix3(this.invMomentOfInertia)
   }
 
+  updateCenter() {
+    const safeFields = {}
+    let safeSum = new THREE.Vector3()
```
pjanik: This could be a `const`. Note that reference is never changed, only underlying object. It's quite a common style in JS to use `const` that actually can be. And it's used within this project, so I'd keep doing it. Not saying that this convention provides much value, but just to be consistent. :wink:

js/plates-model/plate.js
```diff
@@ -79,6 +84,49 @@ export default class Plate extends PlateBase {
...
+    let safeArea = 0
```
pjanik: This is a correct use of `let` in this project. :wink:

js/plates-model/plate.js
```diff
@@ -79,6 +84,49 @@ export default class Plate extends PlateBase {
...
+        })
+        if (safe) {
+          safeFields[field.id] = field
+          safeSum = safeSum.add(field.absolutePos)
+          safeArea += field.area
+        }
+      }
+    })
+
+    if (safeArea < 100000) {
```
pjanik: I usually try to keep such magic numbers as camel case constants at the top of the file.

js/config.js
```diff
@@ -13,6 +13,8 @@ const DEFAULT_CONFIG = {
   playing: true,
   // Save model state every N steps. It can be restored later.
   snapshotInterval: 100,
+  // How many steps between plate centers are recalculated.
+  centerUpdateInterval: 15,
```
pjanik: It might be a matter of taste, but I'd say that values around 60-120 look better actually. Labels move less and seem more stable. Especially in these final moments before plate disappears or in the `circles` test.
Also, not important here, but I'd use `config` for values that Sarah or other authors might be interested in setting (I guess this one is not likely), otherwise just local constants in given file.

js/plates-model/plate.js
ekosmin: Whoops! I actually was changing the reference before, I didn't realize I could mutate Vectors. Fixed now, nice catch

ekosmin:
Responded to your comments and merged, thanks @pjanik!

PR #34, opened 2018-06-08
Improve continent buffers handling

pjanik:
[#157605459]

Make continent-continent collisions more likely. When two continent buffers are colliding, we know that it's not continent-ocean collision case (where we want to spot plates), but rather continent-continent (where we want collision to happen).

### DIFF
--- a/js/plates-model/fields-collision.js
+++ b/js/plates-model/fields-collision.js
@@ -41,13 +41,18 @@ export default function fieldsCollision (bottomField, topField) {
 
   if (bottomField.isOcean) {
     subduction(bottomField, topField)
-    if (bottomField.isContinentBuffer && !topField.isContinent) {
+    if (bottomField.isContinentBuffer && !topField.isContinent && !topField.isContinentBuffer) {
       // Special case when the continent is "trying" to subduct under the ocean. Apply drag force to stop both plates.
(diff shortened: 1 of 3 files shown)


js/plates-model/fields-collision.js
```diff
@@ -41,13 +41,18 @@ export default function fieldsCollision (bottomField, topField) {
...
+      // There's one exception - when both fields are continent buffers, it means continents are about to collide soon.
+      // Don't apply forces in this case, so the orogeny can actually happen.
       applyDragForces(bottomField, topField)
     }
   } else if (bottomField.isContinent) {
     if (topField.isContinent) {
       orogeny(bottomField, topField)
+    } else if (topField.isContinentBuffer) {
+      // Continents are next to each other. They will collide soon. Remove oceanic field to let that happen.
+      topField.alive = false
```
ekosmin: To clarify: does removing the top field here actually change the behavior of the collision? Or just make the collision look better.

pjanik: It does. We remove ocean crust that otherwise would collide with the continent and won't allow orogeny to happen.

pjanik:
Thanks. Not sure, I left it in PT comments. This edge-case doesn't seem to be fixable and I'd say model even should behave like that. I guess it could be another story if necessary.

PR #36, opened 2018-06-11
Making the blank planet wizard load by default

ekosmin:
What do you think of a fix like this for the default wizard, @pjanik? I think the extra reload adds a fraction of a second to the load time, which barely noticeable but not ideal. Perhaps I should be trying to modify the config instead?

### DIFF
--- a/js/components/app.js
+++ b/js/components/app.js
@@ -10,10 +10,13 @@ const authoring = getURLParam('authoring')
 const preset = getURLParam('preset')
 const planetWizard = getURLParam('planetWizard')
 const modelId = getURLParam('modelId')
+const samples = getURLParam('samples')
 
 const App = () => {
   if (authoring) {
     return <Authoring />
+  } else if (samples) {
+    return <IndexPage />
   } else if (preset || modelId || planetWizard) {
     return (
       <Provider simulationStore={simulationStore}>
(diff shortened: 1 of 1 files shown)


pjanik:
What if we just assume that:
1. no URL params open planetWizard
2. `?samples` / `?index` opens index page
3. planetWizard=true can be just ignored

Then no redirect is necessary. Of course, it won't be backward compatible if you use index page, but no one was using that in the curriculum I guess. And it will be backward compatible if you were using planetWizard=true (it won't break anything).

ekosmin:
So I think assumptions 1 and 2 are good. I'm less sure about assumption 3, though, since `planetWizard` gets read in a bunch of different places e.g:

- Changing the text that appears in the 'about' box
- Deciding if the 'reload' button appears in the bottom bar
- Even if a preset is used, you can still set the planet wizard to appear (maybe this feature isn't necessary?)

I couldn't think of an obvious way to keep all these behaviors without explicitly setting `planetWizard=true` somewhere

pjanik:
Good point. Well, you could just change `planetWizard` to false / true in config using some new logic (based on other URL params or lack of them) and I guess all these points would work fine. Although not sure if it's worth the effort, perhaps not at this point.

PR #37, opened 2018-06-18
Cypress Tests

ekosmin:
Hi @pjanik! I'm tagging you since it sounds like you're checking out Cypress tests, but I hope others will look as well. Like I mentioned during the demo, some things didn't work like I expected - I'll leave comments to show where I ran into difficulties


cypress/integration/bottom_bar_spec.js
```diff
@@ -0,0 +1,84 @@
...
+describe('Bottom Bar integration', function () {
+  beforeEach(() => {
+    mount(
+      <Provider simulationStore={simulationStore}>
+        <BottomPanel />
+      </Provider>)
+  })
+
+  it ('Calls the sidebar function', function () {
+    Cypress.component().then((component) => {
```
ekosmin: `Cypress.component()` comes from the [cypress-react-unit-test](https://github.com/bahmutov/cypress-react-unit-test) library, and yields the mounted component. I needed to use a `then` here to hack the `BottomPanel` out of the `Provider` that's wrapping it.

cypress/integration/bottom_bar_spec.js
```diff
@@ -0,0 +1,84 @@
...
+      let bottomPanel = component._reactInternalInstance._renderedComponent._instance.wrappedInstance
+      assert.deepEqual(bottomPanel.state, {
+        sidebarActive: false,
+        fullscreen: false
+      })
+      const spy = cy.spy(bottomPanel, 'toggleSidebar')
+      cy.contains("menu").click()
+      cy.contains("close").click()
```
ekosmin: I really do not know why this happens, but the tests don't pass with a single click here, so I needed to add extra clicks.

cypress/integration/top_bar_spec.js
```diff
@@ -0,0 +1,39 @@
...
+
+  it ('renders correctly', function () {
+    Cypress.component().its('state').should('deep.equal', {
+      shareDialogOpen: false,
+      aboutDialogOpen: false
+    })
+    Cypress.component().then((component) => {
+      const spy = cy.spy(component, 'openAboutDialog')
+      cy.contains('About').click()
+      cy.contains('About').click().then((about) => {
```
ekosmin: Again, I expected to be able to perform a single click and then check the spy in the next line. For some reason, I needed to perform two clicks, *and* wrap the spy check in a `then`.

cypress/utils/cypress_react_test.js
```diff
@@ -0,0 +1,81 @@
+// This file is taken from https://github.com/bahmutov/cypress-react-unit-test
+// The only difference is it references React 15 instead of React 16
```
ekosmin: Like the comment says, I copied the [cypress-react-unit-test](https://github.com/bahmutov/cypress-react-unit-test) library here to get it to work...forking the repo might be smarter, but it's pretty small.

js/components/bottom-panel.js
```diff
@@ -23,7 +23,7 @@ function toggleFullscreen () {
 }
 
 @inject('simulationStore') @observer
-export default class BottomPanel extends PureComponent {
+export default class BottomPanel extends Component {
```
ekosmin: Unrelated change, but the tests were complaining that this isn't a `PureComponent` any more since it has state.

pjanik:
@ekosmin, sorry, it got lost back then, but I was just looking at this PR. It all looks very cool!
@kevinsantos-cc started another branch with Cypress tests:
https://github.com/concord-consortium/tectonic-explorer/tree/cypress-smoke-test
I've manually copied your integration tests there, so we can reuse it.
I didn't copy unit tests and utils, as it all looks more like unit testing in Cypress rather than integration tests. It's definitely cool and nice to know that we can do it. But it turned out that most of our projects use Jest + Enzyme for that, so I wouldn't do it differently here. But it was useful research, thanks!

PR #38, opened 2018-12-14
Add a new option that removes "density| word from Planet Wizard

pjanik:
https://www.pivotaltracker.com/story/show/162387016

New option + text changes.
Also, I've updated test libs, added Enzyme lib, configured it and added the first unit test of a component.

### DIFF
--- a/js/components/sortable-densities.js
+++ b/js/components/sortable-densities.js
@@ -12,10 +12,6 @@ function hueToBackground (hue) {
   return {backgroundColor: 'rgb(' + Math.floor(rgb.r) + ', ' + Math.floor(rgb.g) + ', ' + Math.floor(rgb.b) + ')'}
 }
 
-function minKey (array) {
-  return array.map(plate => plate.id).sort()[0]
-}
-
 const DragHandle = SortableHandle(() => <FontIcon value='menu' className='hamburger-menu' />)
 
 const SortableItem = SortableElement(({plateInfo}) =>
--- a/js/components/sortable-densities.js
+++ b/js/components/sortable-densities.js
@@ -82,7 +82,7 @@ export default class SortableDensities extends Component {
           {
             config.densityWordInPlanetWizard
               ? 'Click and drag to reorder the plate density'
-              : 'Click and drag to reorder the plate that will go below the other plate'
+              : 'Click and drag to reorder the plates'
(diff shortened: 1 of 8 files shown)


js/components/sortable-densities.js
```diff
@@ -77,11 +78,17 @@ export default class SortableDensities extends Component {
...
+          { config.densityWordInPlanetWizard ? 'LOW' : 'ABOVE' }
           <SortableList plateInfos={this.plateInfos} onSortEnd={this.onSortEnd} useDragHandle={false} />
-          HIGH
+          { config.densityWordInPlanetWizard ? 'HIGH' : 'BELOW' }
+        </div>
+        <div className='helper-text'>
+          {
+            config.densityWordInPlanetWizard
+              ? 'Click and drag to reorder the plate density'
+              : 'Click and drag to reorder the plate that will go below the other plate'
```
ekosmin: Is this the desired wording? I'd think something like "Click and drag to order the plates from top to bottom" is more clear, especially when there are more than 2 plates

test/components/sortable-densities.test.js
```diff
@@ -0,0 +1,34 @@
...
+const store = {
+  model: {
+    plates: []
+  }
+}
+
+describe('SortableDensities component', () => {
+  beforeEach(() => {
+    // Restore config as it can be modified by tests.
+    Object.assign(config, originalConfig)
```
ekosmin: To fully restore the config, including new keys, would you want to do something like `Object.assign({}, originalConfig)` instead?

pjanik: Good point, thanks!

js/components/sortable-densities.js
pjanik: I've used wording provided by Amy. I'll double check with her.

pjanik: Amy agreed that this prev version was confusing. She provided another variant that I used in the last commit.

test/components/sortable-densities.test.js
pjanik: That wouldn't work actually, as it's not enough to overwrite local `config` reference. I've added another approach to delete new keys. If we use this pattern more, I guess we should move it to some helper method.

PR #39, opened 2019-01-31
Resolving a series of security issues and updating dependencies

dstrawberrygirl:
A security issue in webpack-dev-server lead to a cascade of dependency updates, the brief summary of which includes:

- Upgrade to webpack 4 - the format of the webpack configuration file has been changed slightly to add a mode key (which I set to development) and module.loaders has been replaced by module.rules.
- json-loader no longer required in webpack configuration, and uglifyjs is now built-in to production builds.
- postcss-loader updated, but postcss-cssnext has been deprecated. Switched to postcss-preset-env as a drop-in replacement as suggested by previous owner of postcss-cssnext
- Autoprefixer is included in postcss-preset-env


webpack.config.js
```diff
@@ -12,8 +12,9 @@ module.exports = {
     path: path.join(__dirname, 'dist'),
     filename: '[name].js'
   },
+  mode: 'development',
```
pjanik: I guess it should depend on `PRODUCTION` env variable (that is set by package.json script).

pjanik:
@dstrawberrygirl, I've just a pushed commit that fixes PostCSS problems. I know that postcss-next isn't recommended anymore, but the new package didn't seem to work. There are no security warnings anyway, so I guess we're fine.

And I think it'd make sense at some point to replace all the components with material-ui instead of trying to fix these issues, as this library seems to be the winner these days.

pjanik:
Somehow this commit didn't get added to this PR.
Here it is (in master too):
https://github.com/concord-consortium/tectonic-explorer/commit/cc1e27d8e576083c81a236b0ecef2405da72eb5b

PR #40, opened 2019-02-12
Snapshot based tests, Cypress drag helper, fake WebGL renderer, TravisCI integration

pjanik:
Changes:
 - Cypress-image-snapshot ingtegration and setup.
 - New Cypress commands (cc @kevinsantos-cc): `waitForSplashscreen`, `waitForSpinner`, `mainCanvasDrag`.
 - Snapshot-based integration tests. They cover PlanetWizard really well + they use sample models to check initial load and model state after some time.
 - Cypress now runs on TravisCI. It required a fake THREEJS.WebGLRenderer implementation. Otherwise, Electron and Chrome were crashing (in a different way, but still).
 - Artifacts are uploaded to S3 (e.g. failed screenshot diffs and generic Cypress screenshots of failed tests).
 - README updated
 - two random bugs fixed

Useful fragment of the README:

> Note that this project heavily depend on GPU/WebGL what makes Cypress tests a bit tricky. TravisCI doesn't provide an access to GPU. That's why there is a separate NPM script `test:cypress:travis` which skips GPU-dependant tests.
> While working locally on new features, you can use regular `npm run test:cypress`. There's also special special `npm run test:cypress:snapshots` script that runs only snapshot-based tests and doesn't stop if image diffing fails for one of them.
> Note that screenshot-based tests should be run using Electron. You can also run these tests in the interactive Cypress mode in headed Chrome (`npm run test:cypress:open`) but screenshots probably will not match e.g. if you have HDPI display. Using dedicated script and Electron lets us avoid these problems.


kevinsantos-cc:
Looks good Piotr!

PR #44, opened 2019-02-19
Earthquakes rendering

pjanik:
Note that this PR is open against https://github.com/concord-consortium/tectonic-explorer/pull/43 as it was based on it.

It adds initial rendering of the earthquakes, following the same shape and style as one used in Seismic Explorer. I've tried to make sure that the code can be reused for volcanoes.

Note that earthquakes are disabled by default and you need to add `earthquakes=true` URL param.

Demo:
http://tectonic-explorer.concord.org/branch/162386261-earthquakes-rendering/index.html?preset=subduction&earthquakes=true

### DIFF
--- a/js/config.js
+++ b/js/config.js
@@ -16,6 +16,8 @@ const DEFAULT_CONFIG = {
   playing: true,
   // If true, the model will show randomly generated earthquakes.
   earthquakes: false,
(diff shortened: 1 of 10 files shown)


pjanik:
@dstrawberrygirl, thanks, I've fixed checkbox issue now, so going to merge it.

PR #45, opened 2019-02-20
Add magnitude and depth to earthquake and handle these properties in rendering code

pjanik:
[#162386261]
[#163915062]

@dstrawberrygirl, further improvements to earthquakes rendering and modeling.
Now, the model stores information about magnitude and depth. Rendering code respects depth, but the magnitude is not handled yet. Rendering requires cleanup and tweaks, but I'm pushing these changes, so you can use depth in the cross-section view.


pjanik:
@dstrawberrygirl, I've just pushed more changes. Rendering is now fixed and earthquakes won't stop just before they disappear if the whole plate is moving.

Also, I've refactored modeling and added new rules. Tried to keep all that in `earthquake.js` helper and added a few URL params (`config` values), so Amy can play with it.

Depth function might require tweaking, but without cross-section, I can't see how good or bad it's working. I guess you could rebase your code on top of that and then use depth value in the cross-section.

js/plates-model/earthquake.js
```diff
@@ -0,0 +1,58 @@
...
+
+  constructor (field) {
+    // Earthquake can be shallow or deep. Shallow are placed around top plate crust, while deep are placed
+    // where two plates are colliding (so we use colliding field elevation as a base).
+    const subductingField = field.colliding
+    let earthquakeElevation
+    if (subductingField) {
+      // Subduction zone
+      const deep = subductingField && random() < 0.5
+      earthquakeElevation = deep ? subductingField.elevation - 0.2 * random() : field.elevation - 0.3 * random()
```
dstrawberrygirl: I'm not sure what the constants of 0.2 or 0.3 represent here

pjanik:
> I think we may need to revisit depth a little more - in the cross section work I have tried plugging the depth directly into the cross section visualization and the quakes appear in very random locations, rather than only in the Lithosphere, as Amy mentioned.

Right, but until I have your cross-section working, I can't really tweak it. As I can't see where they do show up relative to the lithosphere. I think it's just a matter of tweaking some numbers here (actually ones that you asked about in the inline comment).

I hope you noticed it's depth, not an elevation, right? 
So in the cross section, you can't really say:
`drawEarthquake(ctx, f1.earthquakeDepth)`
but you should subtract depth from the top of the crust:
`drawEarthquake(ctx, f1.elevation - f1.earthquakeDepth)`

js/plates-model/earthquake.js
pjanik: It's range of the random depth part.
If the earthquake is shallow, it's placed in `[field.elevation, field.elevation - 0.3]` range.
If the earthquake is deep, it's placed in `[subductingField.elevation, subductingField.elevation - 0.2]` range (around area where two plates collide).

If you look at some comments and config, you should find information that elevation range is between 1 and -3.3 (-3.3 is defined by `config.subductionMinElevation`):
1 is the highest moubtain possible. 0 is the deepest trench. Anything below 0 is subducting plate. -3.3 is the min elevation of the subducting plate and then we remove field that went below this value:

```
  //  - [0, 1] -> [the deepest trench, the highest mountain]
  //  - 0.5 -> sea level
  //  - [config.subductionMinElevation, 0] -> subduction
```

So, 0.2 and 0.3 are some magic params based on this [0, -3.3] range. Most likely they will require tweaking when we can see earthquakes in the cross-section.

dstrawberrygirl: this makes a lot more sense, thanks. I've made changes to the cross section so hopefully it should be a more accurate reflection of these values.

PR #46, opened 2019-02-20
Earthquakes cross-section

dstrawberrygirl:
This initial work will render the earthquakes in cross section view and display them at relative magnitude sizes. The depth calculation may need to be changed at the earthquake level or pending review feedback on how to interpret the data. There is currently an issue with rendering the quakes when using the subduction preset (?preset=subduction&debugCrossSection=true&earthquakes=true) showing them below and to the left of where the top plate meets the subducting plate

### DIFF
--- a/js/plates-view/render-cross-section.js
+++ b/js/plates-view/render-cross-section.js
@@ -1,6 +1,7 @@
 import * as THREE from 'three'
 import config from '../config'
 import magmaSrc from '../../images/magma.png'
+import depthToColor from './earthquake-helpers'
 
 export const OCEANIC_CRUST_COL = '#27374f'
 export const CONTINENTAL_CRUST_COL = '#643d0c'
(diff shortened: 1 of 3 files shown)


js/plates-view/render-cross-section.js
```diff
@@ -122,6 +147,20 @@ function renderChunk (ctx, chunkData) {
   }
 }
 
+function renderChunkEarthquakes (ctx, chunkData) {
+  for (let i = 0; i < chunkData.length - 1; i += 1) {
+    const x1 = chunkData[i].dist
+    const f1 = chunkData[i].field
+
+    // Until we implement depth correctly for earthquakes, place them on the boundary beetween crust and lithosphere
+    const c1 = new THREE.Vector2(x1, f1.elevation - f1.crustThickness)
```
pjanik: I wouldn't try to tweak it that way. I'd use a correct approach (f1.elevation), wouldn't worry about imperfect depth in this PR (as that's not part of it), and then tweak the model.

js/plates-view/render-cross-section.js
```diff
@@ -69,6 +75,24 @@ function drawMarker (ctx, crustPos) {
   ctx.fill()
 }
 
+function drawEarthquake (ctx, xPos, elevation, earthquake) {
+  const earthquakeSize = (1 + Math.ceil(earthquake.magnitude))
+  const strokeWidth = earthquakeSize * 0.1
+  const x = scaleX(xPos)
+
+  // Not ideal depth calculation until actual depth is in the correct range
+  const adjustedDepth = elevation - earthquake.depth
```
pjanik: Why isn't it "ideal" and "adjusted"? I think it's all correct here and shouldn't require any more changes. Now, it's only model tweaking.
I'd say it's just "earthquake elevation" (by elevation I mean anything between 1 and -3.3, as that's the elevation range in our model).

PR #49, opened 2019-02-21
Add volcanoes option

dstrawberrygirl:
Added the option for volcanoes to authoring, to the sidebar, and to the store, similar to the recent changes made to add the Earthquakes to the map. This doesn't yet add volcanoes to the map.


pjanik:
Looks good 👍

PR #50, opened 2019-02-21
Initial implementation of caveat notice

dstrawberrygirl:
The wording is identical for both Earthquakes and Volcanoes so this will show up on screen if either option is currently enabled.


dstrawberrygirl:
Good feedback, tests added!

PR #54, opened 2019-02-27
Erthquakes and volcanic eruptions tests

pjanik:
This PR adds tests related to earthquakes, volcanic eruptions, checks their serialization, and also adds some general improvements to unit testing - loading time should be greatly improved (due to configurable Voronoi sphere grid size).
 
I think the most valuable are tests based on screenshots, and I managed to cover pretty much everything - checkboxes, rendering in planet view, cross-section, and hiding these icons when the model is stopped. I also added some basic unit tests.


pjanik:
One more changed - probabilities depend on the grid size now.

PR #57, opened 2019-03-01
Earthquakes and Volcanic Eruptions in the key

dstrawberrygirl:
Adding Earthquakes and Volcanic Eruptions to the key with a button to toggle the visibility of the key. Each section only appears if its corresponding UI setting is enabled, hence only see the Earthquake key if earthquakes are visible on the map.


pjanik:
It looks much better to me now, thanks! 
I guess it's ready, right?

PR #61, opened 2019-03-06
Key improvements

dstrawberrygirl:
Moved the key for cross-section view from the cross-section to the key
Updated the key UI to group together sections and laid out all pieces in a series of `tbody` elements to ensure all sections lined up when visible.
Updated Cypress screenshots for GPU-based tests


pjanik:
@dstrawberrygirl, I've pushed some small fixes to issues that were mentioned yesterday during demo:

- z-index (make sure key is in front of the cross-section)
- make styling a bit more consistent (e.g. size of the earthquake
  rectangles match cross-section rectangles, etc.)
- set max-height of the key and enable scrolling if necessary

PR #62, opened 2019-03-15
Update earthquake modeling and rendering

pjanik:
This PR addresses Amy's comments:
https://www.pivotaltracker.com/story/show/164541758

- Earthquakes depth is now an absolute value (more like elevation). This ensures that colors always match if two earthquakes are next to each other in the cross-section. Previously, it wasn't that obvious (as depth was relative to field elevation, so mountain on top was pushing them up or down in the cross-section).
- Color scale is updated according to Amy's comments.
- There are fewer shallow earthquakes.
- Key uses color array, so it doesn't require updating next time we update detph to scale function.


js/plates-model/earthquake.js
```diff
@@ -33,13 +34,16 @@ export default class Earthquake {
...
-      const deep = subductingField && random() < 0.5
-      const baseField = deep ? subductingField : field
-      const availableRange = baseField.crustThickness + baseField.lithosphereThickness - MIN_DEPTH
-      const earthquakeElevation = baseField.elevation - MIN_DEPTH - random() * availableRange
-      this.depth = field.elevation - earthquakeElevation
+      // Also, make sure that shallow earthquakes don't show up in trench-areas, as cross section would look confusing.
+      const shallow = (field.elevation - field.crustThickness) > subductingField.elevation & random() < SHALLOW_EQ_PROBABILITY
+      const baseField = shallow ? field : subductingField
+      const availableRange = baseField.crustThickness + (shallow ? 0 : baseField.lithosphereThickness) - MIN_DEPTH
+      this.depth = Math.max(
```
dstrawberrygirl: I'm noticing that some earthquakes in cross-section appear to be almost entirely outside of the lithosphere region, with only a very small overlap. I'm not sure if we need to account for earthquake size in the calculations to account for this so that the center point of an earthquake remains always within that zone - Amy may have thoughts on this.

pjanik: Sometimes it happens because we smooth and average field elevation using a few fields around. If you move cross-section line slightly up or down, the relative position of earthquake and lithosphere might change. It'll be really difficult to improve, so I'd wait for Amy's feedback - if it's a problem or not. Personally, I don't think it's worth the effort, as these differences are small, and on some pictures in the internet earthquakes are pretty spread out.

pjanik: I think this issue has been mostly fixed. Amy noticed the problem too and following commits are trying to address that.

PR #63, opened 2019-06-21
Update UI according to the new spec

pjanik:
Spec: https://app.zeplin.io/project/5cf7c3cecee3f113e909e492/screen/5cf7c4002eb4af5c6e929949
Demo: http://tectonic-explorer.concord.org/branch/166842460-ui-updates/index.html?preset=subduction
(the last version should be available in 10 minutes)


dstrawberrygirl:
Look good! A couple of small changes - `Cross-section` should have lowercase s, and in the Key when Earthquakes are visible, change `Magnitude and Depth` to have lowercase and. Otherwise 👍

pjanik:
Hm, this is how CSS `text-transform: capitalize` works. So, I'll have to go through all the labels and do that by hand. :wink: But if it looks strange otherwise, I guess it makes sense, not a big thing. Thanks.

pjanik:
This change got pretty big, as I had to go through all the labels and fix them + update Cypress screenshots. But I guess it's better not to rely on automatic `capitalize`.

PR #64, opened 2021-02-17
Dependencies update & maintenance

pjanik:
All the dependencies have been updated to the most recent versions available.
Code changes were necessary mostly because of:
1. MobX v3 to v6 migration.
2. THREE.JS r102 to r119 migration.

THREE.JS is not using the most recent version because of the performance bug I've found there. It should be fixed in r126, and I'll update it then. See: https://github.com/mrdoob/three.js/issues/21284

I've updated Cypress config so screenshot-based tests work with its recent version. Also, I've added a new NPM script so it's possible to test screenshots using headed Chrome (there's a note about that in readme too). Snapshots are also updated, as some small issues seem to be fixed there (plate labels colors).

Babel dependencies have been removed. I've used TypeScript compiler instead. It's obviously necessary for TS support, but also the latest `@babel/preset-env` version caused a huge performance drop. Probably something was transpiled in an inefficient way.

I've also removed standard-js linter. I'll try to move to ESLint and our typical config. But I'll do it in a separate PR, as it'll be lots of diff there probably.


package.json
```diff
@@ -56,67 +39,59 @@
...
-    "mobx": "^3.6.2",
-    "mobx-react": "^4.4.3",
+    "lodash": "^4.17.20",
+    "mobx": "^6.1.6",
+    "mobx-react": "^7.1.0",
     "point-in-polygon": "^1.0.1",
-    "postcss": "^6.0.23",
-    "react": "^16.8.3",
-    "react-dom": "^16.8.3",
+    "postcss": "^8.2.6",
```
kswenson, line 54: Isn't this a `devDependency`?

package.json
```diff
@@ -56,67 +39,59 @@
...
-    "copy-webpack-plugin": "^4.6.0",
-    "css-loader": "^2.1.0",
-    "cypress": "^3.3.1",
-    "cypress-image-snapshot": "^3.1.1",
-    "enzyme": "^3.9.0",
-    "enzyme-adapter-react-16": "^1.10.0",
-    "file-loader": "^0.11.2",
-    "fsevents": "^1.2.7",
+    "@svgr/webpack": "^5.5.0",
+    "@wojtekmaj/enzyme-adapter-react-17": "^0.4.1",
```
kswenson, line 69: May want to make a note somewhere to update to the official adapter once it's available.

kswenson (approved):
👍 Nice work -- I skipped the MobX 6, React 17, and Webpack 5 updates on my last round of CLUE dependency updates because I didn't want to deal with some of the issues raised. When I get around to it I may look back at this PR for some ideas. 😉 

One more question: a number of the snapshot PNGs are reported by GitHub UI as being empty files. Is that expected?

dstrawberrygirl (approved):
Looks good, and tested it out locally and it was a clean upgrade from the previous version. Not a blocker, but I saw a lot of high severity warnings in npm audit that went away with a fix - not entirely sure what the root of it was, or whether you get similar warnings, but would be nice to be free of those.

package.json
pjanik: Right, fix in the last commit, thanks.

pjanik:
@dstrawberrygirl, thanks, I've just done the fix and pushed the updated package-lock.

@kswenson,
> One more question: a number of the snapshot PNGs are reported by GitHub UI as being empty files. Is that expected?

Hm, strange, I think it might be a Github bug. If you click "view the file", you can see it. So, I'm assuming it is there. :wink:

PR #65, opened 2021-02-19
Add ESLint, fix all the errors and warnings

pjanik:
I've used ESLint config from starter projects:
https://github.com/concord-consortium/starter-projects/blob/master/.eslintrc.js
with a few additions:
```javascript
  "jsx-quotes": ["error", "prefer-double"],
  "object-curly-spacing": ["error", "always"],
  "array-bracket-spacing": ["error", "never"],
```
- jsx-quotes to be conistent with the main quotes rule
- object-curly-spacing and array-bracket-spacing are set so they match the .eslntrc.js file itself. ;) I have no strong opinion about a particular config, but it's nice to have these rules, as we can call linter with --fix argument and fix all the files.

A more important change is that I've added linting of the test files using cypress and jest eslint plugins. Config is at the bottom and it overrides the main one:

```javascript
    { // rules specific to Jest tests
      files: ["**/*.test.*"],
      env: {
        node: true,
        jest: true
      },
      plugins: ["jest"],
      extends: ["plugin:jest/recommended"],
      rules: {
        "@typescript-eslint/no-require-imports": "off",
        "@typescript-eslint/no-non-null-assertion": "off",
        "jest/no-done-callback": "off"
      }
    },
    { // rules specific to Cypress tests
      files: ["cypress/**"],
      env: {
        node: true,
        "cypress/globals": true
      },
      plugins: ["cypress"],
      extends: ["plugin:cypress/recommended"],
      rules: {
        "@typescript-eslint/no-require-imports": "off",
        "@typescript-eslint/no-non-null-assertion": "off",
        "cypress/no-unnecessary-waiting": "off"
      }
    },
```

95% of the changes in the actual source code were done automatically using `--fix` option.


dstrawberrygirl (approved):
This looks great, and 0 issues in npm audit is a win! nice work :)

PR #66, opened 2021-02-23
Migrate source code to TypeScript

pjanik:
TypeScript migration has been done using https://www.npmjs.com/package/ts-migrate and multiple manual tweaks.
This tool renamed files nicely, provided lots of automatic typing (`any`), and then disabled TS error reporting for around 160 lines. I've actually fixed all of these 160 lines except from one - you could find // @ts-expect in one file.

Unfortunately, the tool has broken formatting in many files and I couldn't find a solution for that. So, I've added more ESLint rules that re-formatted the code automatically +I did manual fixing.

While working on that, I've fixed some things that got broken during previous maintenance - the fullscreen button and earthquakes rendering. That's why Cypress snapshots are updated again.

I'll open a separate PR for more useful TS conversion of some core parts of the model. Also, I might convert tests later too. But this should already let developers write new TS code in a convenient way.


dstrawberrygirl (approved):
This looks great!

PR #67, opened 2021-02-25
Add lots of real TS types and interfaces

pjanik:
This not a conversion script anymore. This PR adds quite a lot of real types added manually.
It covers all the files in ./plate-models, ./stores, some in ./peels, ./interactions and most of the top-level files. 
I haven't worked on React components or 3D view as this seems less useful for new TecRock features. The current state should significantly improve working on new model features and new views.

### DIFF
--- a/js/geo-utils.ts
+++ b/js/geo-utils.ts
@@ -1,7 +1,8 @@
 import * as THREE from "three";
+import { IVector3 } from "./types";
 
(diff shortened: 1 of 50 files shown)


dstrawberrygirl (approved):
This is so much better! great to be rid of so many "any" types, nice work!

PR #68, opened 2021-02-26
Refactor and type state migration, convert tests to TS

pjanik:
I've made serialization more explicit so I could add TS interfaces to describe and document data. 
I had to add small state migration.
Also, tests are now converted to TS, as that was simple.


dstrawberrygirl (approved):
This looks great! 🦖

PR #70, opened 2021-03-03
Add debug mode and plate visibility toggles

pjanik:
[#177183322]
[#176905501]


knowuh (approved):
Looks good. Thank you.

PR #72, opened 2021-03-24
Fix density selector UI and remove canvas outline

pjanik:
This PR fixes Step 4 in the planet wizard:
https://tectonic-explorer.concord.org/branch/master/index.html?planetWizard=true
If you get there, you should notice that densities are draggable, but once dragging stops, they always go back to their initial positions.

Also, there's a tiny fix that removes canvas outline. When user clicks anywhere in the main view or cross-section, a blue outline is added. It doesn't seem useful, so I just removed that.

### DIFF
--- a/js/stores/simulation-store.ts
+++ b/js/stores/simulation-store.ts
@@ -85,6 +85,8 @@ export class SimulationStore {
     if (config.modelId) {
       this.loadCloudModel(config.modelId);
     }
+    // For debugging purposes
+    (window as any).s = this;
   }
 
   // Computed (and cached!) properties.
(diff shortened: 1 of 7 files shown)


js/stores/simulation-store.ts
```diff
@@ -85,6 +85,8 @@ export class SimulationStore {
     if (config.modelId) {
       this.loadCloudModel(config.modelId);
     }
+    // For debugging purposes
```
knowuh, line 88: Did you want to keep this in?  I have no opinion, just checking incase you wanted to remove it.

knowuh (approved):
Looks great.  Checked it out, and it worked fine too.

js/stores/simulation-store.ts
pjanik: Thanks, yes, actually it was intentional. There's something similar in web worker - it exports model instance. It gets useful quite often.

PR #73, opened 2021-03-25
Add basic rock layers

pjanik:
To see these updates, open:
https://tectonic-explorer.concord.org/branch/176906271-rock-layers/index.html?preset=subduction
Let the model run for a bit, and try to draw a cross-section.
You can compare it with the old version:
https://tectonic-explorer.concord.org/?preset=subduction

The cross-section now shows rock layers that are building oceanic or continental crust.
Implemented logic covers:
 - oceanic rifting: basalt and gabbro increases over time)
 - orogeny: rocks get folded and increase their thickness
 - volcanic activity: volcanic rocks added 
 - sedimentation: everything submerged in the ocean is covered with a layer of sediments over time
 - accretionary wedge formation: it happens around the convergent boundary

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,8 +73,8 @@ While working on input data, HSV colors should be used:
   - 0.4 is the deepest possible ocean (0.0 value would be black, so plates couldn't be distinguished from each other)
   - 0.7 is the sea level
   - 1.0 is the highest possible mountain
-  - [0.4, 0.55] is assumed to be oceanic crust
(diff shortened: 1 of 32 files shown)


knowuh (approved):
This looked good to me.  There were many changes that were just changing the spelling of neighbor.

pjanik:
Right, I should have opened a separate PR with that. I was just annoyed with VS Code highlighting it all the time. ;) Thanks for looking at it.

PR #74, opened 2021-03-26
Render the top rock layer in 3D view

pjanik:
This PR lets users change color scheme (using menu) in 3D view and show the top rock layer. Demo:
https://tectonic-explorer.concord.org/branch/177477605-rock-3d/index.html?preset=subduction

Note that this PR nicely predents how to transfer new property from Web Worker to the main thread.
The new array is added in `model-output.ts` and then it's handled in `field-store.ts` while this class is processing post message from the worker. 

I changed `Rock` enum to use numbers, as that lets us transfer values easily using post message transferrable properties (Int16Array). String can't be transported that way.

### DIFF
--- a/js/plates-model/crust.ts
+++ b/js/plates-model/crust.ts
@@ -1,12 +1,14 @@
 import { BASE_OCEANIC_CRUST_THICKNESS, FieldType } from "./field";
 
+// Do not use automatic enum values, as if we ever remove one rock type, other values shouldn't change.
+// It would break deserialization of the previously saved models.
 export enum Rock {
-  Granite = "Gr",
-  Basalt = "Ba",
-  Gabbro = "Ga",
-  MaficRocks = "MaR",
-  AndesiticRocks = "AnR",
-  Sediment = "Se"
+  Sediment = 0,
+  Granite = 1,
+  Basalt = 2,
+  Gabbro = 3,
+  MaficRocks = 4,
+  AndesiticRocks = 5,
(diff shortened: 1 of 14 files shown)


js/stores/field-store.ts
```diff
@@ -1,18 +1,20 @@
...
-  normalizedAge: number | null = null;
+  elevation = 0;
+  normalizedAge = 0;
   boundary = false;
   earthquakeMagnitude = 0;
   earthquakeDepth = 0;
   volcanicEruption = false;
   force = new THREE.Vector3();
   originalHue: number | null = null;
+  rockType: Rock = 0;
```
knowuh, line 17: Again, just for documentation, using `Rock.Sediment` might be nice.

knowuh (approved):
Look good, thanks.  

I learned some things as usual.

I just had a suggestion to use enum types for some code clarity in a few places.

js/plates-model/crust.ts
```diff
@@ -63,6 +65,10 @@ export default class Crust {
     return result;
   }
 
+  get topRockType() {
+    return this.rockLayers[0]?.rock || 0;
```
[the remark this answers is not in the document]
pjanik: Thanks, fixed in the last commit.

js/stores/field-store.ts
pjanik: Thanks, fixed in the last commit.

PR #75, opened 2021-04-07
Render elevation in the 3D view

pjanik:
Demo links:
- default settings - 64 topo shades, flat shading, bump mapping ON: https://tectonic-explorer.concord.org/branch/177509275-elevation-3d/index.html?preset=subduction
- same, but with bump mapping OFF - cleaner / more cartoon-like style: https://tectonic-explorer.concord.org/branch/177509275-elevation-3d/index.html?preset=subduction&bumpMapping=false
- same, but also without flat shading: https://tectonic-explorer.concord.org/branch/177509275-elevation-3d/index.html?preset=subduction&bumpMapping=false&flatShading=false
- “realistic” settings - 256 topo shades, no flat shading, bump mapping ON: https://tectonic-explorer.concord.org/branch/177509275-elevation-3d/index.html?preset=subduction&topoColormapShades=256&flatShading=false

You can compare that with old way of rendering:
https://tectonic-explorer.concord.org/?preset=subduction

The most tricky part is updated shaders and colormap texture being read in the fragment shader. It fixes the initial blurriness of this rendering method. If I only used color attributes per vertex, GPU would linearly interpolate colors. Now the GPU interpolates elevation instead (what is what we actually want), then I'm reading this interpolated elevation in the fragment shader, and finally using it to get correct colormap color.

It works only for continuous colormaps like the topographic one. When rock type coloring is used, values cannot be interpolated (e.g. going from sediments to basalt would result in pixels colored in other rock types that are between these two colors in the color map). So, the color attribute is still used in some cases.

I've tried to add enough comments to make that more or less clear.

### DIFF
--- a/js/peels/sphere/to-cg.ts
+++ b/js/peels/sphere/to-cg.ts
@@ -37,7 +37,7 @@ interface IAttributesResult {
   normals: Float32Array;
   indices: Uint16Array | Uint32Array | Array<number>;
   colors: Float32Array;
-  uvs?: Float32Array;
+  uvs: Float32Array;
 }
 
 /**
@@ -57,6 +57,8 @@ function barycenterVerticesAndFaces(sphere: any, options: IOptions) {
 
   const colors = new Float32Array(indices.length * 4);
 
+  const uvs = new Float32Array(n * 2);
+
   for (let f = 0; f < sphere._Fields.length; f += 1) {
     const field = sphere._Fields[f];
 
@@ -66,14 +68,21 @@ function barycenterVerticesAndFaces(sphere: any, options: IOptions) {
 
     const color = options.colorFn.call(field);
 
-    positions[f * 3 + 0] = cos(fφ) * cos(fλ); // x
-    positions[f * 3 + 2] = cos(fφ) * sin(fλ); // z
-    positions[f * 3 + 1] = sin(fφ); // y
+    const x = cos(fφ) * cos(fλ);
+    const z = cos(fφ) * sin(fλ);
+    const y = sin(fφ);
+
+    positions[f * 3 + 0] = x;
+    positions[f * 3 + 1] = y;
+    positions[f * 3 + 2] = z;
 
     colors[f * 3 + 0] = color.r;
     colors[f * 3 + 1] = color.g;
     colors[f * 3 + 2] = color.b;
     colors[f * 4 + 3] = color.a != null ? color.a : 1.0;
+
+    uvs[f * 2 + 0] = 0.5 + Math.atan2(z, x) / (2 * Math.PI);
+    uvs[f * 2 + 1] = 0.5 - Math.asin(y) / Math.PI;
   }
 
   // normals are exactly positions, as long as radius is 1
@@ -83,7 +92,8 @@ function barycenterVerticesAndFaces(sphere: any, options: IOptions) {
     positions,
     normals,
     indices,
-    colors
+    colors,
+    uvs
   };
 }
 
--- a/js/plates-view/plate-mesh-fragment.glsl
+++ b/js/plates-view/plate-mesh-fragment.glsl
@@ -4,7 +4,9 @@
 
 // --- CUSTOM:
 varying vec4 vColor;
-varying float vNormElevation;
+varying float vHidden;
+varying float vColormapValue;
+
 uniform sampler2D colormap;
 uniform float HIDDEN_FIELD_ALPHA_VAL;
 // ---
@@ -32,6 +34,7 @@ uniform float opacity;
 #include <lights_pars_begin>
 #include <lights_phong_pars_fragment>
 #include <shadowmap_pars_fragment>
+
 // #include <bumpmap_pars_fragment>
 // Use custom bump map helpers that use bumpScale attribute instead of uniform.
 #ifdef USE_BUMPMAP
@@ -75,6 +78,7 @@ uniform float opacity;
 
 #endif
 // --- CUSTOM
+
 #include <normalmap_pars_fragment>
 #include <specularmap_pars_fragment>
 #include <logdepthbuf_pars_fragment>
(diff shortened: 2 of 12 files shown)


js/peels/sphere/to-cg.ts
```diff
@@ -66,14 +68,21 @@ function barycenterVerticesAndFaces(sphere: any, options: IOptions) {
...
+    positions[f * 3 + 0] = x;
+    positions[f * 3 + 1] = y;
+    positions[f * 3 + 2] = z;
 
     colors[f * 3 + 0] = color.r;
     colors[f * 3 + 1] = color.g;
     colors[f * 3 + 2] = color.b;
     colors[f * 4 + 3] = color.a != null ? color.a : 1.0;
+
+    uvs[f * 2 + 0] = 0.5 + Math.atan2(z, x) / (2 * Math.PI);
```
knowuh, line 84: Just making a note that I don't completely understand this UV mapping, but I don't know if I need to understand it. 

To me it looks like we are creating a UV map with one entry for for each field(which are X,Y,Z locations mapped over a sphere.) and setting the value of the UV map to a value computed from the X,Y,Z components of the field.

js/plates-view/plate-mesh-fragment.glsl
```diff
@@ -84,9 +91,21 @@ void main() {
...
+  if (vHidden == 1.0) {
+    // Hide pixel.
+    diffuseColor = vec4(0.0, 0.0, 0.0, 0.0);
+  } else if (vColor.a > 0.5) {
+    // Use specific color provided in color attribute. For example plate boundary color.
+    // Note that rendering only for alpha > X (e.g. 0.75 or 0.5) makes this color line thinner.
+    diffuseColor *= vec4(vColor.r, vColor.g, vColor.b, 1.0);
+  } else {
+    // Use the default colormap if vColor is transparent.
+    diffuseColor *= texture2D(colormap, vec2(0.5, vColormapValue));
```
knowuh, line 105: I am assuming that `vColor` is the vertex color and the `colorMap` sampler indexes into the UV map.

Just learning. :D

js/plates-view/plate-mesh-vertex.glsl
```diff
@@ -28,10 +34,14 @@ varying vec3 vViewPosition;
...
-    vBumpScale = vertexBumpScale;
-    // ---
+
+  // --- CUSTOM:
+  vColor = color;
+  vBumpScale = vertexBumpScale;
+  vHidden = hidden;
+  vColormapValue = colormapValue;
+  // ---
+
```
knowuh, line 44: Again, I am not 100% sure how glsl vertex shaders work, but it seems like you are assigning values to the `varying` fields from the attributes(input?), which I guess get passed into the next step in the shader pipeline.  

I guess I am most puzzled by `transformed` assignment on 62, I don't see where that variable is being defined, nor `objectNormal`, those must defined in one of the `#include`s?  

The idea I think is to display the vertex by elevation towards its normal (away from the center of the sphere).

knowuh (approved):
Nice to see this, thanks for sharing it.

Looks good to me, there are a few things I don't 100% understand, but everything seems sensible.

js/peels/sphere/to-cg.ts
pjanik: > To me it looks like we are creating a UV map with one entry for for each field(which are X,Y,Z locations mapped over a sphere.) and setting the value of the UV map to a value computed from the X,Y,Z components of the field.

Yes, that's true.

In general in the OpenGL world, UV coordinates map each 3D vertex of a mesh to some 2D (X, Y) coordinates within the range [0, 1]. These 2D coordinates are used to tell shaders which point of a 2D texture should be used for a given mesh vertex.

When you map a rectangular plane in 3D, that's obviously pretty easy. One corner would be (0,0), another (1,0), another (1,1) and (0,1), and that's all. When we deal with a sphere, these calculations get more complex, as it's pretty close to a map projection in fact.

Also, these UV coordinates are **not** used by the colormap texture that I've added in this PR. They're used by a bump map texture (and could be used by other, standard textures that we could add to THREE.JS material).

js/plates-view/plate-mesh-fragment.glsl
pjanik: > I am assuming that vColor is the vertex color

Yes, vColor is interpolated vertex color attribute. Note that we're in a fragment shader. This shader is run per each pixel on the screen. So, when you transition from vertex shader (run per each vertex only) to fragment (run per each pixel), vertex attributes are interpolated in a smart way.

>  the colorMap sampler indexes into the UV map.

Actually not. UV coords are not used here. colormap is a texture that has dimensions 1x<number_of_shades>. When you read its value using texture2D, you need to provide coords within range [0,1]. X is not important, so that's why I just use 0.5. And Y is the shade that we want to use. This is set using `vColormapValue`. It's an interpolated `colormapValue` vertex attribute that we set in JavaScript code. For a topo colorscale, it'd be normalized elevation ([0, 1] range so it works with texture).

js/plates-view/plate-mesh-vertex.glsl
pjanik: > Again, I am not 100% sure how glsl vertex shaders work, but it seems like you are assigning values to the varying fields from the attributes(input?), which I guess get passed into the next step in the shader pipeline.

Yes. I mentioned that earlier, but this shader is run per each mesh vertex. The next fragment shader is run per each pixel. There much more pixels than vertices. So, when we assign an attribute to `varying`, it gets linearly interpolated. OpenGL will interpolate nicely these vertex attributes to fill all the space of the final, rendered triangle.

> I guess I am most puzzled by transformed assignment on 62, I don't see where that variable is being defined, nor objectNormal, those must defined in one of the #includes?

Yes, exactly, it's somewhere in the included pieces. This is a pretty hacky way to write shaders. I'm reusing THREE.JS PhongMaterial shader, and making just minimal custom changes. That way, I can reuse all the features of this material without rewriting the whole shader. Also, when THREE.JS and PhongMaterial get updated, it should be relatively easy to update this shader too.

> The idea I think is to display the vertex by elevation towards its normal (away from the center of the sphere).

Yes. That's a fast way to apply elevation to each vertex. Another approach is to simply modify position attributes.

Actually, I use both ways, depending on `config.flatShading`. 
The fast way done in a shader doesn't let us fix normals (as normals need to be calculated using a few vertices around and there's no access to that in shader code). So, geometry will be modified, but normals would be outdated, and lighting/shading will be incorrect (it'd actually look like there's no elevation). But if we use `flatShading` mode, normals are not necessary, and then this method is good enough.

If we don't use `flatShading`, I use the slow way - I update position attributes in JS code, and then recalculate normals all the time. It ensures that lighting is also correct, but is a bit slower.

pjanik:
@knowuh, I've left some explanations. If it's still unclear, and you're interested, I'm happy to meet and answer more questions. 😉

PR #81, opened 2021-05-14
Add trench modeling

pjanik:
I've ended up with relatively simple implementation.

`subduction.progress` is replaced with more general `bendingProgress` attribute, which is defined for each field. That way plate can bend even before subduction happens. It's done in `propagateBending` method. There are also some tweaks to wedge to cover unwanted break between the wedge and the top plate (it could create an impression that it's the deepest part of the ocean).


knowuh (approved):
Looks good.

PR #82, opened 2021-05-20
Update rendering of magma around divergent boundaries

pjanik:
Old cross-section:
Screenshot 2021-05-20 at 15 40 10
⬇️ ⬇️ ⬇️
New cross-section:
Screenshot 2021-05-20 at 15 39 20

The magma animation is done in a very simple way, but it works. An obvious use case where animation would be twice as fast is if the user was looking at two divergent boundaries at the same time. But it almost never happens in practice, as the cross-section line is too short.

Also, there are some small tweaks of rock colors requested by Stephanie.

### DIFF
--- a/js/plates-view/render-cross-section.ts
+++ b/js/plates-view/render-cross-section.ts
@@ -8,6 +8,7 @@ import { OCEANIC_CRUST_COL, CONTINENTAL_CRUST_COL, LITHOSPHERE_COL, MANTLE_COL, 
 import { IChunkArray, IEarthquake, IFieldData } from "../plates-model/get-cross-section";
 import { SEA_LEVEL } from "../plates-model/field";
 import { rockColor } from "../colormaps";
+import { UPDATE_INTERVAL } from "../plates-model/model-output";
 
 export interface ICrossSectionOptions {
   rockLayers: boolean;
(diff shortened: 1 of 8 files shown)


js/plates-view/render-cross-section.ts
```diff
@@ -65,6 +66,57 @@ function drawMagma(ctx: CanvasRenderingContext2D, top: THREE.Vector2) {
...
+  const animationProgress = Math.pow(animationStep / (animationStepsCount * 0.5), 0.5);
+  
+  ctx.save();
+
+  ctx.beginPath();
+  ctx.rect(Math.min(p1XScaled, p3XScaled), p1YScaled + (1 - animationProgress) * clipRectHeight, clipRectWidth, animationProgress * clipRectHeight);
+  ctx.clip();
+
+  const gradient = ctx.createLinearGradient(0, p1YScaled, 0, p3YScaled);
+  gradient.addColorStop(0, "#fc3c11");
```
knowuh, line 102: It might be nice to name some of these color RGB values and put them all in one place with descriptive (color) names. 

RGB values are very hard for me to interpret, so this isn't actually about what the colors represent, although, yes, where that makes sense to do, it would even be better.

knowuh (approved):
Looks good, thanks.

js/plates-view/render-cross-section.ts
pjanik: Sure, I'll start doing that incrementally when I stumble upon some hex values. Most of the colors are in colormap.ts and cross-section-colors.ts, but right, definitely not everything. I'll try to clean it up. 👍

PR #85, opened 2021-06-11
Tweak plate division and add new presets to test it easier

pjanik:
Changes to plate division algorithm. I've tried something much fancier (the new plate was adding new fields using BFS with priority queue where fields were weighted by distance from mountains and center of the new plate), but it didn't produce better results and the project team was fine with the more random approach.

[#178022770]


knowuh (approved):
looks good.

PR #86, opened 2021-06-14
Update rock layers rendering, mountain roots forming + small refactoring

pjanik:
This PR changes the way how rock layers are rendered. You can compare the screenshots below.

The main difference is that rock layers shouldn't have jagged edges anymore. CS rendering will interpolate between various layers. However, this requires all the rock layers to have a defined, specific order. So, sediments will always be the topmost layer, then rhyolite, etc, down to granite or gabbro. Only then I can connect rock layers between neighboring fields. Otherwise, if layers are totally mixed, we'd get crossing lines. But this constraint seems to be fine, there's no use case where layers should have a different order. 

Another changes in this PR:
- Mountain roots are formed in a different way to avoid the impression that volcanic rocks are "sinking" and getting lower than the initial crust elevation. Stephanie and Amy ware asking for that.
- Sediments are removed when volcanic rocks are added. This was requested by Chris.
- Small performance optimizations

old 1
new 1
old 2
new 2

### DIFF
--- a/js/plates-model/crust.ts
+++ b/js/plates-model/crust.ts
@@ -2,15 +2,17 @@ import { BASE_OCEANIC_CRUST_THICKNESS, FieldType } from "./field";
 
 // Do not use automatic enum values, as if we ever remove one rock type, other values shouldn't change.
 // It would break deserialization of the previously saved models.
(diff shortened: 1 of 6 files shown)


knowuh (approved):
Looks good, thanks!

PR #88, opened 2021-06-28
Fix densities setup and some of the plate store observable properties

pjanik:
https://www.pivotaltracker.com/story/show/178693575

```
// Recreate platesMap to update all its observers (e.g. SortableDensities component).
this.model.platesMap = new Map(this.model.platesMap);
```

This line was causing the problem. The whole map was recreated unnecessarily and breaking some observers. This line was added to work around other issues in fact. Plate density wasn't updating its observers, as PlateStore#denisty property didn't have a default value. Once I set it to 0, this property really became observable and I could delete the problematic line.

Why initial values are necessary? I mentioned that on Slack:

If someone is using MobX + TS actively, something to watch out for (it took me hours to fix a bug caused by that):
```
class MyClass {
  @observable myProperty: number;
```
…will not work. myProperty won’t be observable even though you’ve just added the annotation. To fix it just assign a default value:
```
class MyClass {
  @observable myProperty = 0;
```
Or even:
```
class MyClass {
  @observable myProperty: number | undefined = undefined;
```
I guess under the hood annotations are transformed into something like makeObservable(<initialValue>) and obviously, it’s required for them to work. I bet TypeScript might simply strip out the whole line if there’s no initial value, as this kind of declaration useless for pure JS engine. In the end we’re ending up with a property that is not observable and won’t trigger re-renders and all the things that you expect from observable MobX properties.


knowuh (approved):
Thanks for the MobX observer lesson. 

Looks good.  Thank you

PR #90, opened 2021-06-30
Add new rock types (sandstone, shale, limestone), refactor rock properties handling and fix some smaller issues

pjanik:
The main change is that now we have 3 more rock types: sandstone, shale, limestone. They cover continents. Shale and limestone are visible only on the continental shelf.

Screenshot 2021-06-29 at 15 15 30
Screenshot 2021-06-30 at 16 09 38

Demo:
https://tectonic-explorer.concord.org/branch/178449561-new-rock-types/index.html?preset=subduction
https://tectonic-explorer.concord.org/branch/178449561-new-rock-types/index.html?preset=continentalCollision1
https://tectonic-explorer.concord.org/branch/178449561-new-rock-types/index.html?preset=islandCollision2
https://tectonic-explorer.concord.org/branch/178449561-new-rock-types/index.html?&planetWizard=true

I've updated continentalCollision1 preset - I've made continent boundaries smoother, so these new rocks can be visible there. There's also a small refactoring of rock properties, I hope it's much cleaner now. Plus one random rendering bug fix.

There's some magic related to imports... If I import `Rock` type from `./rock-properties` in `get-cross-section.ts` file, the model gets broken. There are no plates visible. If I import the same enum from `./crust`, things work fine. Absolutely weird, but that's why I'm re-exporting Rock in the `crust.ts` file. I've added a comment about it too. I hope I have more time to investigate it.


js/types.ts
```diff
@@ -1,3 +1,17 @@
+export enum Rock {
```
knowuh, line 1: I know that you are having a mystery with importing `Rock` from various places not behaving consistently.  But I think its maybe confusing that it is defined in multiple places like this. 

Maybe you forgot to get rid of this definition while you were experimenting?

knowuh (approved):
Looks good to me. 

I am intrigued by the import/export of `Rock` definitions you describe here.  Looking forward to hearing that resolution once you work it out, which I am sure you will eventually.

js/types.ts
pjanik: Right, thanks, it should not be there. But I added it to types.ts when I already had this problem and I was testing if it helps - it doesn't. I've just removed this duplicated declaration in the last commit.

knowuh: Looks good.

pjanik:
@knowuh,

> I am intrigued by the import/export of Rock definitions you describe here. Looking forward to hearing that resolution once you work it out, which I am sure you will eventually.

FYI: it was caused by circular dependencies between constants in different TS files. It somehow worked in the web browser, but it was failing in Node.js. That's how I found the problem. It's been fixed in these two commits:
- https://github.com/concord-consortium/tectonic-explorer/commit/bda5ed68f6ca89809678315b0d0b42a21a9bd36b
- https://github.com/concord-consortium/tectonic-explorer/commit/b039d2600f63473a9e6d2eeecedc0062cf958b5b

kswenson:
>FYI: it was caused by circular dependencies between constants in different TS files. It somehow worked in the web browser, but it was failing in Node.js. That's how I found the problem. It's been fixed in these two commits:

If I'm reading this right, it looks like you still have a circular dependency in the sense that `crust.ts` imports from `field.ts` and vice-versa.

pjanik:
Right, I should have been more specific. I've explained it better here:
https://github.com/concord-consortium/tectonic-explorer/pull/93#discussion_r666342802

PR #91, opened 2021-07-01
Maintenance - update THREE.js, use `pointerevents`, add new Cypress tests

pjanik:
THREE.js update was quite important, as there were some breaking changes, that also included shader code. It'll make future updates easier. I've added `pointerevents` support, as Orbit Controls from THREE.js moved to that. I had to update Cypress helper methods.

Most of the dependencies are now updated to the very last version (except for d3 libs that are now published to NPM in a silly way). 

There's also a new Cypress test that tests both interactions and actually its own dragging helpers. It was useful when I was debugging `pointerevents` issue, so I left them.

Cypress is now updated to v7.6.0, but it seems the build still randomly gets stuck and there's no output for 10mins. I'll try to investigate it more.


knowuh (approved):
Looks good

knowuh:
Code looks good, but tests are failing.

pjanik:
> Code looks good, but tests are failing.

Right, I mentioned it here:

> Cypress is now updated to v7.6.0, but it seems the build still randomly gets stuck and there's no output for 10mins. I'll try to investigate it more.

But it seems the new test had 100% failure rate on Travis, it wasn't even random. I removed it for now.

package.json
```diff
@@ -60,17 +60,17 @@
...
   },
   "devDependencies": {
     "@svgr/webpack": "^5.5.0",
-    "@types/d3-color": "^2.0.1",
+    "@types/d3-color": "^3.0.1",
     "@types/d3-hsv": "^0.1.4",
-    "@types/d3-interpolate": "^2.0.0",
-    "@types/d3-scale": "^3.2.2",
+    "@types/d3-interpolate": "^3.0.0",
+    "@types/d3-scale": "^4.0.0",
```
kswenson, line 73: Generally speaking, `@types` major version numbers track the corresponding library's major version numbers. If you're sticking with `d3-color: ^2.0.0`, for instance, you should stick with `@types/d3-color: ^2.0.0`, otherwise you run the risk of having type definitions that don't correspond to the version of the code you're using. As for the underlying issue, I've run into the published-as-ESM-modules problem with d3 and other libraries as well, and agree that it's annoying.

test/model-serialization.test.ts
```diff
@@ -22,7 +22,7 @@ function initFunc(plates: Record<number, Plate>) {
 let modelImgData: ImageData | null = null;
 
 beforeAll(done => {
-  (global as any).getModelImage("testModel.png", (imgData: ImageData) => {
+  (self as any).getModelImage("testModel.png", (imgData: ImageData) => {
```
kswenson, line 25: Having used the `(global as any)` trick in test code myself, I'm curious about the reason for this change. Is there some advantage I should be aware of?

pjanik: After I updated a bunch of dependencies, something was complaining about `global`. I don't remember now if it was ESlint or TS (or probably TS-Jest).

package.json
pjanik: Thanks, I've forgotten to revert versions of @types packages. I've just fixed that in the last commit.


Re d3, ESM and other breaking changes - I cannot count the number of times D3 libs did that to me. It's definitely the lib that I wasted the most time on trying to update versions or make them work in a new environment (ofc not in this project). With all the respect to its developers, maintainers, and their good ideas - no idea why I keep using them. :wink:

kswenson: I've also run into it with a number of Sindre Sorhus' libraries, so d3 aren't the only ones.

pjanik:
OK, it seems that these updates actually fixed the old flaky tests. Just the new one wasn't working great. Issues pointed out by Kirk are also fixed, so I'm going to merge this PR.

PR #92, opened 2021-07-02
Use rock pattern images in the cross-section view, refactor helpers related to colors

pjanik:
Live demo of the patterns:
https://tectonic-explorer.concord.org/branch/178726485-cs-patterns/index.html?preset=subduction
I’ve added a new URL param that can disable patterns and just leave updated solid colors - crossSectionPatterns=false:
https://tectonic-explorer.concord.org/branch/178726485-cs-patterns/index.html?preset=subduction&crossSectionPatterns=false
The key has been updated too. I’ve also added the green border around magma blobs. 
I've tried to clean up code related to colors and color utils (converting between all the formats used by this app).


package.json
```diff
@@ -34,7 +34,9 @@
       "<rootDir>test/setupTests.js"
     ],
     "moduleNameMapper": {
-      "\\.(css|less|scss)$": "identity-obj-proxy"
+      "\\.(css|less|scss)$": "identity-obj-proxy",
+      "\\.svg$": "<rootDir>/__mocks__/svgMock.js",
+      "\\.(jpg|jpeg|png|gif|eot|otf|webp|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/__mocks__/fileMock.js"
```
kswenson, line 39: It seems that some of our projects (e.g. CLUE) use `__mocks__` consistently for this purpose, while others use `identity-object-proxy` consistently, and other use a mix of both (like this). Do you have a rule-of-thumb for when to prefer one over the other?

test/plate-division-merge.test.ts
```diff
@@ -4,8 +4,6 @@ import config from "../js/config";
 import Plate from "../js/plates-model/plate";
 import { compareModels } from "./serialization-test-helpers";
 import * as seedrandom from "../js/seedrandom";
-import dividePlate from "../js/plates-model/divide-plate";
-import Field from "../js/plates-model/field";
 
 // Increase this paramter so it uses to real-world value (setup-tests.js sets a tiny value for performance reasons).
```
kswenson, line 8:
```suggestion
// Increase this parameter so it uses to real-world value (setup-tests.js sets a tiny value for performance reasons).
```

js/colors/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+
+// Color object used internally by 3D rendering. RGB values are within [0, 1] range.
+export type RGBAFloat = { r: number; g: number; b: number; a: number; };
+
+const toFloatRGB = 1 / 255;
+export function d3RGBToRGBAFloat(d3Rgb: RGBColor): RGBAFloat {
+  return { r: d3Rgb.r * toFloatRGB, g: d3Rgb.g * toFloatRGB, b: d3Rgb.b * toFloatRGB, a: d3Rgb.opacity };
+}
+
+export function cssColToRGBAFloat(cssColorSpecifier: string): RGBAFloat {
```
kswenson, line 11:
```suggestion
export function cssColorToRGBAFloat(cssColorSpecifier: string): RGBAFloat {
```
You can afford the extra two characters. 😉

kswenson (approved):
👍 Looks good -- code is definitely clearer with the changes. Just a few thoughts on naming and one refactoring suggestion that I leave to your discretion.

Also, there's one image filename with a space (`oceanic sediment.png`), which I generally try to avoid.

js/colors/rock-colors.ts
```diff
@@ -0,0 +1,104 @@
...
+};
+
+const ROCK_MAIN_COLOR_RGBA_FLOAT = (() => {
+  const result: Partial<Record<Rock, RGBAFloat>> = {};
+  Object.keys(ROCK_PATTERN).forEach((key) => {
+    const rock = key as unknown as Rock;
+    result[rock] = cssColToRGBAFloat(ROCK_PATTERN[rock].mainColor);
+  });
+  return result as Record<Rock, RGBAFloat>;
+})();
```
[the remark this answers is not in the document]
pjanik: Thanks. I tried to implement it in a suggested way in the last commit. I just didn't make fgbaFloat optional since I know it's actually always provided.

js/colors/cross-section-colors.ts
```diff
@@ -1,12 +1,17 @@
...
 export const CONTINENTAL_CRUST_COL = "#643d0c";
-export const LITHOSPHERE_COL = "#666";
-export const MANTLE_COL = "#4b1e01";
-export const SKY_COL_1 = "#4375be";
-export const SKY_COL_2 = "#c0daeb";
+
+export const LITHOSPHERE_COL = "#5e505d";
+export const MANTLE_COL = "#531a1e";
+export const SKY_COL_1 = "#4275be";
+export const SKY_COL_2 = "#bcd6e8";
```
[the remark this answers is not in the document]
pjanik: Done.

js/colors/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+  return d3RGBToRGBAFloat(rgb(cssColorSpecifier));
+}
+
+export function RGBAFloatToHexNumber(rgbF: RGBAFloat): number {
+  return Math.pow(2, 16) * Math.round(rgbF.r * 255) + Math.pow(2, 8) * Math.round(rgbF.g * 255) + Math.round(rgbF.b * 255);
+}
+
+
+export function RGBAFloatToCssCol(c: RGBAFloat) {
```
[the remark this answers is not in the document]
pjanik: Fixed.

package.json
pjanik: My understanding is that they're not really interchangeable. `identity-object-proxy` works nicely with CSS modules, as all the named imports are changed to strings and we can even use that in tests. For example:
```
import css from "./styles"

<div className={css.header}>
```
becomes
```
<div className="header">
```




It be nice if it worked for other assets like images, but I've tried and it doesn't:
Screenshot 2021-07-05 at 10 34 48

I guess it's because assets use default imports. That's why I'm using custom files/svg mocks. But if there's a way to use `identity-object-proxy` for everything, I'd happy to do it.

test/plate-division-merge.test.ts
pjanik: Fixed.

js/colors/utils.ts
pjanik: Done ;)

pjanik:
@kswenson, thanks for the comments. I think I've addressed almost all of them in the last commit.

> Also, there's one image filename with a space (oceanic sediment.png), which I generally try to avoid.

Generally I agree and I wouldn't create a file with whitespace myself. But based on my previous experience, it's super practical to keep file names provided by Michael. Then, when he updates the assets, I can just copy-paste them and don't think about mapping filenames. In this particular case, adding underscore could easily be overlooked either by me or especially by some other developer who makes this update in the future. Maybe I'll ask Michael to generally avoid whitespace in filenames in the future and if he changes this one, I'll update it too. Webpack will inline it or change it to some random name anyway.

js/plates-view/planet-view.ts
```diff
@@ -7,14 +7,15 @@ import ForceArrow from "./force-arrow";
...
+import { topoColor } from "../colors/topographic-colors";
+import { RGBAFloatToHexNumber } from "../colors/utils";
 import getThreeJSRenderer from "../get-threejs-renderer";
 
 import "../../css/planet-view.less";
 
 // Mantle color is actually blue, as it's visible where two plates are diverging.
 // This crack should represent oceanic ridge.
-const MANTLE_COLOR = rgbToHex(topoColor(0.40));
+const MANTLE_COLOROR = RGBAFloatToHexNumber(topoColor(0.40));
```
kswenson, line 18:
```suggestion
const MANTLE_COLOR = RGBAFloatToHexNumber(topoColor(0.40));
```
May have overshot the mark just a smidge. :wink:

package.json
kswenson: @eireland ran into this recently in [portal-report #344](https://github.com/concord-consortium/portal-report/pull/344) where we added the
```
"\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "identity-obj-proxy"
```
to fix an error with PNG import, but we only did it that way because that's the way the styles were being imported. In that limited case the `identity-object-proxy` seemed to work, but it sounds like that's not always the case. Given your experience it sounds like `identity-object-proxy` is preferred for styles because it works better with CSS Modules (although this is less important in projects that aren't using CSS Modules, like CLUE), while `__mocks__` is preferred for assets because it works in some situations where `identity-object-proxy` does not (although those limitations apparently don't affect all projects, e.g. portal-report). Thanks for the explanation.

kswenson (approved):
👍 Changes look good -- just one extra "OR" that snuck in. 😉

pjanik:
I've fixed MANTLE_COLOROR typo, so I'm assuming other comments are addressed and it's fine to merge. Thanks for the review!

PR #93, opened 2021-07-08
Fix circular dependencies, serialization and serialization tests, add test helpers and an example integration test

pjanik:


js/plates-model/crust.ts
```diff
@@ -1,10 +1,8 @@
 import config from "../config";
 import { random } from "../seedrandom";
-import { BASE_OCEANIC_CRUST_THICKNESS, BASE_CONTINENTAL_CRUST_THICKNESS, FieldType } from "./field";
+import { FieldType } from "./field";
```
kswenson, line 3: There still appear to be circular imports between `field.ts` and `crust.ts`.

pjanik: I think there are many circular dependencies of that type in this app (and probably most of our apps). This seems okay to me and I think it's unavoidable in some cases. Quite a common pattern here is to have reference to the parent object in the child object. While it's not perfect, it helps a lot. For example, field properties are greatly based on their parent tectonic plate properties - rotation, angular velocity, etc. In this case, we cannot avoid having `import { Plate } from "./plate"` in `field.ts` and `import { Field } from "./field"` in `plate.ts`.

The problem I was describing was a bit different. We had a set of constants A, B, C and X, Y, Z in two different files. And they're evaluated on the module load like:
```
export const A = X + Y;
export const B = Z;
```
and
```
export const X = A + B;
export const Y = X + A;
```
If it was in one file, TS would complain that some of the constants are used before they're defined. But it went unnoticed as definitions were spread between files. Also, there were some functions involved that made things even more confusing. I've fixed that by moving all the related constants and functions to one file and fixing problems detected by TS (in this commit: https://github.com/concord-consortium/tectonic-explorer/pull/93/commits/c8cace4f706965cc3999de7cf22f2361042b7cbd).

pjanik:
@kswenson, should I assume you've already reviewed it? I've forgotten to add a description, but probably the commit names are the best description, as it's a set of half-related changes.

js/plates-model/model-test-helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export function getField(model: Model, plateId: number, fieldId: number) {
+  return model.getPlate(plateId)?.fields.get(fieldId);
+}
+
+export function getFieldElevation(model: Model, plateId: number, fieldId: number) {
+  return getField(model, plateId, fieldId)?.elevation;
+}
+
+export function isFieldSuducting(model: Model, plateId: number, fieldId: number) {
```
kswenson, line 14:
```suggestion
export function isFieldSubducting(model: Model, plateId: number, fieldId: number) {
```

test/integration/subduction.test.ts
```diff
@@ -0,0 +1,36 @@
+import Model from "../../js/plates-model/model";
+import config from "../../js/config";
+import presets from "../../js/presets";
+import { getFieldElevation, getFieldRockLayers, runModelFor } from "../../js/plates-model/model-test-helpers";
+
+// Increase this paramter so it uses to real-world value (setup-tests.js sets a tiny value for performance reasons).
```
kswenson, line 6:
```suggestion
// Increase this parameter so it uses to real-world value (setup-tests.js sets a tiny value for performance reasons).
```

test/integration/subduction.test.ts
```diff
@@ -0,0 +1,36 @@
...
+let modelImgData: ImageData | null = null;
+
+beforeAll(done => {
+  (self as any).getModelImage("../public/" + presets.subduction.img, (imgData: ImageData) => {
+    modelImgData = imgData;
+    done();
+  });
+});
+
+test("subudction model is loaded correctly and volcanoes show up on the top plate", () => {
```
kswenson, line 18:
```suggestion
test("subduction model is loaded correctly and volcanoes show up on the top plate", () => {
```

test/serialization-test-helpers.ts
```diff
@@ -3,6 +3,28 @@ import Plate from "../js/plates-model/plate";
 import Field from "../js/plates-model/field";
 import Subplate from "../js/plates-model/subplate";
 
+// Helper that compares two provided values. When values are numeric, it uses .toBeClose assertion.
+// It handles only one level of recursion to avoid issues with circular dependencies.
+export function expectValuesToBeClose(a: any, b: any, recursive = true) {
```
kswenson, line 8: Clever!

kswenson (approved):
👍 Looks good -- just a few typos

PR #94, opened 2021-07-09
Fix model time calculations, move some of the integration tests to Cypress, add more examples

pjanik:
The first commit fixes model time calculations. I've just noticed it randomly while working on something else. The second one moves the subduction Jest test to Cypress. There are two main reasons for that:
- it works faster in Chrome / Electron
- @tejal-shah is familiar with Cypress

@tejal-shah, so you might be interested in this second commit. The easiest way to develop these tests is to:
- load one of the samples from: https://tectonic-explorer.concord.org/branch/master/index.html?samples
- then open the side menu and select "Log Field Data" interaction
- open the web browser console
- click click somewhere in the planet view to log a field data - you can check field ID, plate ID, elevation, and various properties that you might want to test against. It's pretty easy to explore that in the browser console.
- I've added helpers that should be pretty straightforward to use in `model-test-helpers.ts`. They're used in example tests too. `getField` will return an object identical to one logged by "Log Field Data".

These tests will still be pretty fragile, but less than screenshots. I'd try to mostly use assertions like "greaterThan", "closeTo", and avoid using equality check. Also, I'd try to limit the number of points/fields we check, so it's easier to update the test when necessary. I think we can add a few more and let's see how it works in practice.


kswenson (approved):
👍 Looks good -- This is causing me to reconsider the way I think about test environments. Based on my CLUE experience I had been assuming that jest == fast and cypress == slow, but this suggests that it's more about the nature of the CLUE cypress tests (e.g. visiting multiple pages) than an inherent characteristic of the cypress testing environment. I do prefer the jest expect syntax to the chai syntax, but I suppose I could learn to live with that. 😉

pjanik:
I think your assumptions are generally true. Or at least mine are the same. 😉 Cypress also seems to have a longer warmup/start time. However, these tests are untypical as they run hundreds of computationally intensive model steps. I think that Chrome is just faster than Node.js and here it starts to matter. Plus, Tejal might write some of them, so Cypress should be easier in this case. I also prefer Jest syntax, but I bet our QA team is used to Cypress and its default syntax.

tejal-shah (approved):
Thank you @piotr for sharing this example. I've approved this PR and will start writing some tests based on this hopefully sometime next week. 
In the meantime, would you be able to tell how close the closeTo() needs to be to pass?

pjanik:
> In the meantime, would you be able to tell how close the closeTo() needs to be to pass?

There's a second argument to this function that can be used to specify delta. I don't think there's one answer to that. It's up to you, range of the value you're checking, and the particular test. When I was testing elevation that ranges from 0 to 1.5 (more or less), I was using delta = 0.1. I guess I could use a bigger one too, or even stick to `greaterThan` / `lessThan`.

PR #96, opened 2021-07-14
Fix cross-section calculations

pjanik:
This PR fixes (or at least greatly improves) issues described here:
https://www.pivotaltracker.com/story/show/178804156

Changes:
1. I've improved a bit `debugCrossSection` mode to show plate number for each rock column.
2. Some of the cases I've traced were related to the fact that `.dist` property of cross-section fields was not calculated correctly (after some plate chunks got merged together). It's probably confusing, but each field has three properties: distStart, dist, and distEnd. distStart and distEnd describe how wide is a given field in the cross-section view. `dist` is something that is finally used by the rendering code. Each field is reduced to a single point, as we don't want to render columns. Insteady, we find a center of each field and connect them. So, `dist` will usually equal the center between `distStart` and `distEnd`. But the first point of a given plate will use `distStart` and the last will use `distEnd` to ensure that plate has the correct width and there are no gaps between plates.
3. Other cases causing gaps between plates were related to the fact that rendering code was not handling correctly plate chunks that were made of a single point. I've added a code that splits single point chunks into two points and that's handled correctly.

I know that get-cross-section.ts file might really hard to understand and read. The naming of various properties isn't the best for sure, and I guess I should try to simplify some logic. It took me many hours with a notebook to trace the issues described above. But at the same time, I don't want to redo this whole logic, as generally, it works and I don't see an easy way to simplify it. Also, we have the last 2 days of development time. But a bigger refactoring of this file is definitely on my to-do list during the next dev cycle.

### DIFF
--- a/js/plates-model/get-cross-section.ts
+++ b/js/plates-model/get-cross-section.ts
@@ -31,6 +31,7 @@ export interface IMagmaBlobData {
 
 export interface IFieldData {
   id: number;
+  plateId: number | string;
   elevation: number;
   crustThickness: number;
   rockLayers: IRockLayerData[],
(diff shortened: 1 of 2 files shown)


kswenson (approved):
👍 Looks good -- better you than me. 😉

PR #97, opened 2021-07-14
Reimplement plate draw tool, make it simpler and maintain continental shelfs

pjanik:
https://www.pivotaltracker.com/story/show/178804174

The previous implementation was more complicated and it wasn't setting up correctly the continental shelf while the continent was being erased. I've tried to simplify the implementation and fix this issue.


kswenson (approved):
👍 LGTM -- definitely simpler and easier to follow. Seems reasonable to me.

PR #98, opened 2021-07-15
Make time counter hidden by default, add URL param to show it

pjanik:
[#178415983]

A tiny PR that adds a new URL param.


kswenson (approved):
👍

PR #99, opened 2021-07-15
Make metamorphism configurable, fix and update authoring page

pjanik:


js/components/authoring.tsx
```diff
@@ -43,19 +45,16 @@ const VIEW_OPTIONS: Option[] = [
 const SKIPPED_OPTIONS: Option[] = ["authoring", "planetWizard", "planetWizardSteps", "sidebar", "preset", "modelId", "densityWordInPlanetWizard"];
 
 // All the options manually defined in various sections.
-const CUSTOM_OPTIONS: Option[] = [];
-CUSTOM_OPTIONS.concat(MAIN_OPTIONS, VIEW_OPTIONS, SKIPPED_OPTIONS).map(opt => typeof opt === "string" ? opt : opt[0]);
+const CUSTOM_OPTIONS: Option[] = [].concat(MAIN_OPTIONS as any, VIEW_OPTIONS as any, SKIPPED_OPTIONS as any)
```
kswenson, line 48: Would this work as a means of eliminating the `any`s?
```suggestion
const CUSTOM_OPTIONS: Option[] = [...MAIN_OPTIONS, ...VIEW_OPTIONS, ...SKIPPED_OPTIONS]
```

js/components/sidebar-menu.tsx
```diff
@@ -52,13 +52,15 @@ export default class SidebarMenu extends BaseComponent<IProps, IState> {
   toggleVelocities: any;
   toggleVolcanicEruptions: any;
   toggleWireframe: any;
+  toggleMetamorphism: any;
```
kswenson, line 55: Not that it matters much, but isn't the type here just `() => void` for the `toggle...` functions and then `(value: any) => void` for the `change...` functions?

kswenson (approved):
👍 Looks good -- just a couple comments on types

js/components/sidebar-menu.tsx
pjanik: Thanks, I've added these types.

js/components/authoring.tsx
pjanik: It helped actually, thanks. :) Weirdly, even without the `.map()` call TS wasn't complaining about types in this case, but it's a nicer approach anyway.

pjanik:
I've addressed all the comments, so merging this PR. Thanks!

PR #100, opened 2021-07-15
Improve loading performance

pjanik:
The most time-consuming thing is the creation of the `Grid` singleton instance that includes `VoronoiSphere` calculations. Voronoi Sphere is used to find a field that is closest to a specified position. It's useful for collision detection and a few other core parts of the model. This PR does two things:
 - Creates `Grid` instance in an optimal moment both in web worker and the main thread.
 - Sets an optimal Voronoi sphere field number based on the number of model fields. This number will be 5 times bigger than the number of model fields. The default grid has around 10000 fields, so the Voronoi sphere will have around 50k. The previous value was hardcoded to 200k which seemed to be too much. I've tested the model both with 200k and 50k fields and couldn't see a significant difference.

### DIFF
--- a/js/config.ts
+++ b/js/config.ts
@@ -133,10 +133,10 @@ const DEFAULT_CONFIG = {
     "wireframe",
     "save"
   ],
-  // Used by the grid model and defines quality of the collision detection. Shouldn't be changed in the real model.
-  // It's useful to decrease this value in tests that don't care about precision, as the model will initialized way
-  // faster.
-  voronoiSphereFieldsCount: 200000,
+  // Voronoi sphere is used by the grid model and defines quality of the collision detection. It shouldn't be changed 
+  // in the real model. It's useful to decrease this value in tests that don't care about precision, as the model will 
+  // initialized way faster. `undefined` value means that model will use a preferred value based on the grid size.
+  voronoiSphereFieldsCount: undefined,
   // Display key expanded by default
   key: true,
   // Show a time counter that displays model time in million years.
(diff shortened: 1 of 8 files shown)


js/config.ts
```diff
@@ -133,10 +133,10 @@ const DEFAULT_CONFIG = {
     "wireframe",
     "save"
   ],
-  // Used by the grid model and defines quality of the collision detection. Shouldn't be changed in the real model.
-  // It's useful to decrease this value in tests that don't care about precision, as the model will initialized way
-  // faster.
-  voronoiSphereFieldsCount: 200000,
+  // Voronoi sphere is used by the grid model and defines quality of the collision detection. It shouldn't be changed 
+  // in the real model. It's useful to decrease this value in tests that don't care about precision, as the model will 
+  // initialized way faster. `undefined` value means that model will use a preferred value based on the grid size.
```
kswenson, line 138:
```suggestion
  // be initialized way faster. `undefined` value means that model will use a preferred value based on the grid size.
```

test/integration/model-serialization.test.ts
```diff
@@ -4,7 +4,7 @@ import config from "../../js/config";
 import Plate from "../../js/plates-model/plate";
 import { compareModels } from "../serialization-test-helpers";
 
-// Increase this paramter so it uses to real-world value (setup-tests.js sets a tiny value for performance reasons).
+// Increase this parameter so it uses to real-world value (setup-tests.js sets a tiny value for performance reasons).
```
kswenson, line 7:
```suggestion
// Increase this parameter so it uses a real-world value (setup-tests.js sets a tiny value for performance reasons).
```

kswenson (approved):
👍 LGTM

PR #102, opened 2021-07-28
Adding automated tests for tectonic explorer 2.0 features

tejal-shah:


js/plates-model/model-test-helpers.ts
```diff
@@ -11,8 +11,16 @@ export function getFieldElevation(model: Model, plateId: number, fieldId: number
   return getField(model, plateId, fieldId)?.elevation;
 }
 
-export function isFieldSubducting(model: Model, plateId: number, fieldId: number) {
-  return getField(model, plateId, fieldId)?.subduction?.progress || 0 > 0;
+export function isFieldUnderneathSubducting(model: Model, plateId: number, fieldId: number) {
+  return getField(model, plateId, fieldId)?.subductingFieldUnderneath?.subduction?.progress || 0 > 0;
```
pjanik, line 15: I think that's not working fully as intended. I guess you meant:
`(getField(model, plateId, fieldId)?.subductingFieldUnderneath?.subduction?.progress || 0) > 0;`
Since you're comparing with 0, there's a small difference in practice (you'd usually return undefined or a number instead of boolean), but still.

cypress/integration/basic/simulation-rules/continent-ocean-collision.js
```diff
@@ -0,0 +1,95 @@
...
+    rockLayers = modelHelper.getFieldRockLayers(model, 0, 2458);
+    expect(rockLayers?.[0].rock).to.equal("Oceanic Sediment");
+    expect(rockLayers?.[1].rock).to.equal("Basalt");
+    expect(rockLayers?.[2].rock).to.equal("Gabbro");
+  });
+
+  it("checks subduction of left plate of the convergent boundary", () => {
+
+    //Check Subduction
+    expect(modelHelper.isFieldUnderneathSubducting(model, 1, 967)).to.be.greaterThan(0);
```
pjanik, line 75: Once `isFieldUnderneathSubducting` is fixed, I'd probably just check if it's true. The name of the function suggests it returns true or false.

cypress/integration/basic/simulation-rules/continent-ocean-collision.js
```diff
@@ -0,0 +1,95 @@
...
+    // Right plate - continental shelf
+    rockLayers = modelHelper.getFieldRockLayers(model, 1, 967);
+    expect(rockLayers?.[0].rock).to.equal("Oceanic Sediment");
+    expect(rockLayers?.[1].rock).to.equal("Basalt");
+    expect(rockLayers?.[2].rock).to.equal("Gabbro");
+  });
+
+  it("runs model for 50 million years", () => {
+    modelHelper.runModelFor(model, 50); // million years
+  });
```
pjanik, line 60: There are no assertions in this test. Personally, I'd just execute this command in the test that makes some assertions related to it.

cypress/integration/basic/simulation-rules/continent-ocean-collision.js
```diff
@@ -0,0 +1,95 @@
...
+
+    // Left plate - continent, ocean
+    expect(modelHelper.getFieldElevation(model, 0, 2773)).to.be.closeTo(0.55, delta);
+    expect(modelHelper.getFieldElevation(model, 0, 2458)).to.be.closeTo(0.03, delta);
+
+    // Right plate - ocean
+    expect(modelHelper.getFieldElevation(model, 1, 967)).to.be.closeTo(0.03, delta);
+  });
+
+  it("checks rock layers of left plate at the convergent boundary", () => {
```
pjanik, line 33: This might be personal/opinionated, but I tend to think that `it` refers to the tested subject. In this case "Continent Ocean Collision model". So, I'd probably say something like:
`it("sets rock layers of left plate at the convergent boundary correctly", () => {`
In your case, `it` seems to refer to the test itself. But if you prefer to organize tests that way, I don't think it matters to me. :wink: I can also see that some of the tests in this repo use a similar convention already.

pjanik (changes requested):
Generally looks to me, thanks!
I've left a few small inline comments. I think the only one that might matter is missing `( )` in one of the model test helpers.

tejal-shah:
@pjanik , I've now addressed all 4 comments left by you.

pjanik (approved):
Thanks, it looks good!

PR #103, opened 2021-09-20
Add initial rock key draft

pjanik:
[#179579510]

This PR adds all the rock key assets, necessary components, and event handlers. It's mostly missing data and styling.
To see the added changes:
1.  Go to https://tectonic-explorer.concord.org/branch/179579510-rock-key/index.html?preset=subduction.
2. Click on Draw a Cross-section from the top.
3. Click and drag anywhere on the planet to draw a cross-section.
A legend with rock types should appear on the right.
Clicking on one of the rock types expands that section with more info on the rock type.
Clicking on another rock type contracts the previously selected section, and expands the new section.
Ocean and Sky are not clickable
rock type
.


mklewandowski:
I see I was added as a reviewer here.  Could someone clarify what this PR is?  I've seen this project just once in a meeting so I'm not sure I can add anything useful at this point.  Much of this looks like assets, but could someone help me understand what the "rock key" is?  Is there a UI spec I can look at?  What is the goal in having me review this (is there something I should be looking for or at specifically?)?  Thanks!

js/colors/rock-colors.ts
```diff
@@ -110,3 +110,27 @@ export const getRockCanvasPattern = (ctx: CanvasRenderingContext2D, rock: Rock)
...
+export const OTHER_GRAY_LIGHT = "#F7F7F7";
+
+
+
+
+
+
+
+
+
```
mklewandowski, line 136: Are these extra spaces intentional?  Seems like they can be removed.

mklewandowski (approved):
👍 LGTM.  I added one comment for removing unneeded whitespace.  Also curious if this is from an updated spec as there are likely places where the background and text colors result in color contrasts that do not meet accessibility standards.
image

PR #120, opened 2021-11-01
Extend boundary utilities to support boundary assignment dialogs

kswenson:
Extends the existing boundary utilities with some functions required by the dialog code:
- `highlight-boundary-segment.ts` renamed to `boundary-utils.ts`
- `getBoundaryInfo()` added to `boundary-utils.ts`
- `getBoundaryInfo()` removed from `model-worker.ts`
- `isPolarCap` added to `PlateStore`


pjanik (approved):
Looks good! 👍

PR #122, opened 2021-11-01
Rock Sample tool bug fixes

pjanik:
This PR fixes two bugs found during QA.

https://www.pivotaltracker.com/story/show/177652612/comments/227721882:
> If the key is manually closed and then the rock sample is then taken, should we expect the key to reopen? currently it doesn't and as per the description it seems like it should.

https://www.pivotaltracker.com/story/show/180051337/comments/227722170:
> With take sample enabled, if the cross section is closed and reopened, we lose the take sample setting and the rock sample marker is no longer displayed but even with this normal cursor we are able to take the rock samples and the key updated accordingly. This seems like a bug.


kswenson (approved):
👍 LGTM

PR #124, opened 2021-11-02
Set hot spots / force vectors based on the boundary type

pjanik:
This PR adds the math necessary to set force vectors using boundary type info. It might seem weird to keep fields in the `BoundaryInfo` instead of plates, but they provide more data than just the `plateId` (=> absolute position useful to calculate forces later). At the same time, they're defining `plateId` too.

### DIFF
--- a/js/stores/helpers/boundary-utils.ts
+++ b/js/stores/helpers/boundary-utils.ts
@@ -1,6 +1,6 @@
 import { toSpherical } from "../../geo-utils";
 import getGrid from "../../plates-model/grid";
-import { IBoundaryInfo } from "../../types";
+import { BoundaryType, IBoundaryInfo } from "../../types";
 import FieldStore from "../field-store";
 import ModelStore from "../model-store";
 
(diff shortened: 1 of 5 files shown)


kswenson (approved):
👍 Looks good! There's no telling how long it would have taken me to figure out those vector force calculations. 😉 

I've added a commit to my branch which eliminates the `observer()`s and adds a TODO comment. I'll let you decide whether to what order to merge the PRs in.

js/stores/helpers/boundary-utils.ts
```diff
@@ -82,35 +82,86 @@ export function unhighlightBoundary(field: FieldStore) {
...
+      const lon1 = latLon1.lon + kLonDiff;
+      const latAvg = 0.5 * (lat0 + lat1);
+      const newPos0 = toCartesian([latAvg, lon0]);
+      const newPos1 = toCartesian([latAvg, lon1]);
+      const rotationAxis0 = newPos0.clone().normalize();
+      const rotationAxis1 = newPos1.clone().normalize();
+      const trueNorth0 = getNorthVector(latAvg, lon0);
+      const trueNorth1 = getNorthVector(latAvg, lon1);
+      const force0 = type === "convergent" ? trueNorth0.applyAxisAngle(rotationAxis0, 0.5 * Math.PI) : trueNorth0.applyAxisAngle(rotationAxis0, -0.5 * Math.PI);
+      const force1 = type === "convergent" ? trueNorth1.applyAxisAngle(rotationAxis1, -0.5 * Math.PI) : trueNorth1.applyAxisAngle(rotationAxis1, 0.5 * Math.PI);
```
kswenson, line 147: I just noticed you could DRY up these lines a bit by moving the ternary to the rotation angle argument.

pjanik: Thanks, I just did it, so I'll push it as a part of another PR.

PR #126, opened 2021-11-02
Calculate boundary type

pjanik:
[#179577787]

This PR adds calculations of the boundary type using existing hot spots and boundary orientation. I've used the dynamic approach, as it's reasonably easy with the orientation already calculated.

The nice side effect of that sometimes you can get information whether the given boundary is convergent or divergent even when you click on it for the very first time (if you set up neighbouring plates correctly).


kswenson (approved):
👍 Looks good -- I'll put in the fix for the longitudinal boundary assignment under planet rotation in a subsequent PR.

PR #127, opened 2021-11-03
Fix longitudinal boundary assignment under rotation

kswenson:
This PR improves our heuristic for determine which plate is on which side of a longitudinal boundary segment. ~~It works _most_ of the time, but can still give incorrect results when the longitude values wrap between the boundary and the center of the plate. With our existing presets this code gives the correct results on all longitudinal boundaries except the 4/5 boundary in the uneven five-plate preset with the globe rotated. @pjanik may have a better idea for how to address this.~~

I figured out a better approach.

### DIFF
--- a/js/stores/helpers/boundary-utils.ts
+++ b/js/stores/helpers/boundary-utils.ts
@@ -81,7 +81,7 @@ export function unhighlightBoundary(field: FieldStore) {
   }
 }
 
-export function getBoundaryInfo(field: FieldStore, otherField: FieldStore): IBoundaryInfo | undefined {
+export function getBoundaryInfo(field: FieldStore, otherField: FieldStore, model: ModelStore): IBoundaryInfo | undefined {
   const polarCapField = field.plate.isPolarCap
                           ? field
                           : otherField.plate.isPolarCap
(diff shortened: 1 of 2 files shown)


js/stores/helpers/boundary-utils.ts
```diff
@@ -114,12 +114,22 @@ export function getBoundaryInfo(field: FieldStore, otherField: FieldStore): IBou
...
-    // Order fields from west to east.
-    fields = platePos.lon < otherPlatePos.lon
-              ? [otherField, field]
-              : [field, otherField];
+    const { lat, lon } = toSpherical(field.absolutePos);
+    const fieldRotationAxis = field.absolutePos.clone().normalize();
+    const westTestVec = getNorthVector(lat, lon).setLength(0.5).applyAxisAngle(fieldRotationAxis, 0.5 * Math.PI);
+    const eastTestVec = getNorthVector(lat, lon).setLength(0.5).applyAxisAngle(fieldRotationAxis, -0.5 * Math.PI);
+    const westTestField = model.topFieldAt(field.absolutePos.add(westTestVec));
+    const eastTestField = model.topFieldAt(field.absolutePos.add(eastTestVec));
```
pjanik, line 122: `.add` adds the other vector in place. In this case, it modified `field.absolutePos`. AFAIR it's not a problem, as absolutePos is a getter that returns a new vector each time. But I've been trying to always .clone() it first, just to be safe. For example, if we decide to cache .absolutePos one day (probably would make sense given how many times it's requested).

pjanik (approved):
Looks good. I've tested it with 2 plates and it actually works really nice. If you set a convergent boundary on one side and rotate to the other, it'd show a divergent one there. Even though it's the same boundary. It switches from convergent to divergent exactly at north and south poles.

js/stores/helpers/boundary-utils.ts
```diff
@@ -114,12 +114,22 @@ export function getBoundaryInfo(field: FieldStore, otherField: FieldStore): IBou
...
-    const platePos = toSpherical(field.plate.center);
-    const otherPlatePos = toSpherical(otherField.plate.center);
-    // Order fields from west to east.
-    fields = platePos.lon < otherPlatePos.lon
-              ? [otherField, field]
-              : [field, otherField];
+    const { lat, lon } = toSpherical(field.absolutePos);
+    const fieldRotationAxis = field.absolutePos.clone().normalize();
+    const westTestVec = getNorthVector(lat, lon).setLength(0.5).applyAxisAngle(fieldRotationAxis, 0.5 * Math.PI);
+    const eastTestVec = getNorthVector(lat, lon).setLength(0.5).applyAxisAngle(fieldRotationAxis, -0.5 * Math.PI);
```
pjanik, line 120: Not sure if it matters, but note that the whole globe has radius 1.0. So, this vector is relatively VERY long. :) I guess in some rare cases, if plates are small enough, it could point at the plate that we don't expect here. I'll test if it works well with length = 0.1 and push the change if it does.

kswenson: Yeah, I thought it might be better to reduce the size but I wasn't sure how much. 0.1 sounds better.

PR #128, opened 2021-11-03
Assign boundary buttons

pjanik:
1. This PR fixes a bug in base-interaction-manager. When event handlers were added to `this.view.domElement`, mouse up was not detected in Assign boundary type interaction (it was blocked by the modal). So, it was locking the rotation.
2. I've added two interaction buttons at the top in the Assign boundary type step. I know they're not in Michael's mock, but I think Michael didn't have time yet to design the Planet Wizzard yet. And these buttons will be necessary for continent drawing, etc. Here, I think they provide a bit of info on what to do, we're consistent with other steps, and it might be easier for users to find a way to rotate the planet and assign the boundary at the back. It's almost zero work (except for debugging 1. that should be useful anyway :wink:), so it's fine if we remove / change them when they're redesigned. 

UPDATE 11/5/201: 
After running it by the team, they decided they want to:
1. Lock the rotation completely in the Boundary Type step, even when the user is dragging the black background. 
2. Add smooth animation when the camera is reset.
3. If possible, add URL parameter that would actually enable rotation during the Boundary Type step, so they can make A/B testing.

I've just pushed a few more commits that implement all that. The smooth transition will be used also when the model is running and in the cross-section view.


kswenson (approved):
👍 Code looks good -- I think the behavior is an improvement, but the spec says:

>The “Draw Force Vectors” and “Rotate Camera” buttons have both been removed from this step. The ability to rotate/zoom the planet will no longer be available during this step.

so you should run this by the team.

pjanik:
> 👍 Code looks good -- I think the behavior is an improvement, but the spec says:
> 
> > The “Draw Force Vectors” and “Rotate Camera” buttons have both been removed from this step. The ability to rotate/zoom the planet will no longer be available during this step.
> 
> so you should run this by the team.

Ops, that's somewhere in PT? Sure, I'll ask on Slack.

pjanik:
@kswenson, I've pushed a few more commits and updated the PR description, so it reflects the current state and the project team requirments.

js/stores/simulation-store.ts
```diff
@@ -238,6 +242,10 @@ export class SimulationStore {
     }
   }
 
+  @action.bound setplanetCameraLocked(value: boolean) {
```
kswenson, line 245:
```suggestion
  @action.bound setPlanetCameraLocked(value: boolean) {
```

kswenson (approved):
👍 Looks good -- I like the rotation animation! Just one typo.

PR #129, opened 2021-11-03
Make boundary type dialog backdrop transparent

pjanik:
Following the specs and I think it makes it easier to notice arrows being added to the model.


kswenson (approved):
👍 Looks good!

PR #131, opened 2021-11-04
Use the last version of rock patterns and update naming

pjanik:
I just copied files directly from Michael's google drive folder and kept -cs and -key prefixes so it's clear which file is used where. I think only two files were actually updated/added.

[#179943345]


kswenson (approved):
👍 LGTM

PR #132, opened 2021-11-04
Add rock patters to the globe view

pjanik:
Copying my message from Slack as it might be useful if I ever want to revisit this feature:

> At this point, I’m pretty sure that it’s almost impossible to implement it 100% correctly because of some specifics how GPU work. I’ll try to explain that as simple as possible (might not be necessary for most of the people here, but maybe useful even for me if I ever go back to this issue and forget what was the problem ;)). You can imagine that each vertex in our 3D mesh is a model field. 3 vertices create a triangle, and all triangles together create a 3D globe we can see. We can specify which pattern texture should be used for each vertex of this triangle, but we’re not able to specify which texture should be used for all the pixels between these vertices (=> triangle “body”). GPU does it automatically - it takes each of the triangle vertices and interpolates linearly all the pattern values to have some values for all the pixels that lie inside the triangle. And this approach works well when the neighbouring vertices have similar pattern values like one vertex uses pattern = 3 and another pattern = 4. All the values between 3 and 4 can be rounded to either 3 and 4, so all the pixels between two mesh vertex would have a correct pattern. But if you imagine that neighboring vertices use pattern 3 and 10, the automatic interpolation will result in patterns 3,4,5,6…,9,10. So, we’ll have a full rainbow of patterns between two model fields. You can see that in the attached screenshot if you take a look at volcanoes on the continent - they’re surrounded by narrow stripes of patterns that should not be there. If we could order all the patterns in a way that two patterns are next to each other each time they might be next to each other in the model, we could workaround that. But I’ve tried and it’s impossible - Rhyolite is often surrounded by Sandstone, Shale, or Limestone, and so on. There are too many combinations.

> Everything would work great if we could tell GPU to use nearest neighbor interpolation instead of the linear one, but that’s just impossible. I’ve checked OpenGL specs, docs, looked for similar issues. The only thing I could do was to use no interpolation at all. But then GPU takes value for the whole triangle from one of the three vertices. That’s why the current visualization that looks a bit squarish and sometimes it might not match the underlying model perfectly. This is visible especially with the rock picker - it doesn’t work precisely and sometimes you might be clicking one color and see a different one in the rock key. We’d have this problem even in the perfect scenario, but probably much less visible. On a positive note, it still looks better than the old visualization. :wink:

> I think the only way to improve all that would be to completely change our 3D mesh and represent each model field with a full triangle. But that’s an enormous amount of work that and creates many other issues too.

Re code itself - I hope the comments explain why the GLSL lines look that weird.

### DIFF
--- a/js/plates-view/plate-mesh-fragment.glsl
+++ b/js/plates-view/plate-mesh-fragment.glsl
@@ -1,6 +1,6 @@
 // Slightly modified Phong shader taken from THREE.ShaderLib:
 // https://github.com/mrdoob/three.js/blob/0c51e577afd011aea8d635db2eeb9185b3999889/src/renderers/shaders/ShaderLib/meshphong_frag.glsl.js
-// It supports alpha channel in color attribut (vec4 is used instead of vec3) + a few custom features like 
+// It supports alpha channel in color attribut (vec4 is used instead of vec3) + a few custom features like
 // hiding vertices, colormap texture, and so on. Custom code is always enclosed in // --- CUSTOM comment.
 
 // --- CUSTOM:
@@ -9,6 +9,9 @@ varying float vHidden;
 varying float vColormapValue;
 
 uniform sampler2D colormap;
+// Keep array length equal to number of rock patterns.
+uniform sampler2D patterns[7];
+uniform bool usePatterns;
 // ---
 #define PHONG
 uniform vec3 diffuse;
@@ -102,6 +105,22 @@ void main() {
     // Use specific color provided in color attribute. For example plate boundary color.
     // Note that rendering only for alpha > X (e.g. 0.75 or 0.5) makes this color line thinner.
     diffuseColor *= vec4(vColor.r, vColor.g, vColor.b, 1.0);
+  } else if (usePatterns == true) {
+    if (round(vColormapValue) == 0.0) {
+      diffuseColor *= texture2D(patterns[0], vec2(vUv.x * 100.0, vUv.y * 100.0));
+    } else if (round(vColormapValue) == 1.0) {
+      diffuseColor *= texture2D(patterns[1], vec2(vUv.x * 100.0, vUv.y * 100.0));
+    } else if (round(vColormapValue) == 2.0) {
+      diffuseColor *= texture2D(patterns[2], vec2(vUv.x * 100.0, vUv.y * 100.0));
+    } else if (round(vColormapValue) == 3.0) {
+      diffuseColor *= texture2D(patterns[3], vec2(vUv.x * 100.0, vUv.y * 100.0));
+    } else if (round(vColormapValue) == 4.0) {
+      diffuseColor *= texture2D(patterns[4], vec2(vUv.x * 100.0, vUv.y * 100.0));
+    } else if (round(vColormapValue) == 5.0) {
+      diffuseColor *= texture2D(patterns[5], vec2(vUv.x * 100.0, vUv.y * 100.0));
+    } else if (round(vColormapValue) == 6.0) {
+      diffuseColor *= texture2D(patterns[6], vec2(vUv.x * 100.0, vUv.y * 100.0));
+    }
   } else {
     // Use the default colormap if vColor is transparent.
     diffuseColor *= texture2D(colormap, vec2(0.5, vColormapValue));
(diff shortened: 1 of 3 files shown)


js/plates-view/plate-mesh-fragment.glsl
```diff
@@ -102,6 +107,34 @@ void main() {
     // Use specific color provided in color attribute. For example plate boundary color.
     // Note that rendering only for alpha > X (e.g. 0.75 or 0.5) makes this color line thinner.
     diffuseColor *= vec4(vColor.r, vColor.g, vColor.b, 1.0);
+  } else if (usePatterns == true) {
+    // This doesn't look convincing, but unfortunatelly that's the only way to do it in GLSL (=> version supported by
```
kswenson, line 111:
```suggestion
    // This doesn't look convincing, but unfortunately that's the only way to do it in GLSL (=> version supported by
```

js/plates-view/plate-mesh-vertex.glsl
```diff
@@ -1,6 +1,6 @@
 // Slightly modified Phong shader taken from THREE.ShaderLib.
 // https://github.com/mrdoob/three.js/blob/0c51e577afd011aea8d635db2eeb9185b3999889/src/renderers/shaders/ShaderLib/meshphong_vert.glsl.js
-// It supports alpha channel in color attribut (vec4 is used instead of vec3) + a few custom features like 
+// It supports alpha channel in color attribut (vec4 is used instead of vec3) + a few custom features like
```
kswenson, line 3:
```suggestion
// It supports alpha channel in color attribute (vec4 is used instead of vec3) + a few custom features like
```

js/plates-view/plate-mesh-fragment.glsl
```diff
@@ -1,14 +1,19 @@
 // Slightly modified Phong shader taken from THREE.ShaderLib:
 // https://github.com/mrdoob/three.js/blob/0c51e577afd011aea8d635db2eeb9185b3999889/src/renderers/shaders/ShaderLib/meshphong_frag.glsl.js
-// It supports alpha channel in color attribut (vec4 is used instead of vec3) + a few custom features like 
+// It supports alpha channel in color attribut (vec4 is used instead of vec3) + a few custom features like
```
kswenson, line 3:
```suggestion
// It supports alpha channel in color attribute (vec4 is used instead of vec3) + a few custom features like
```

js/plates-view/plate-mesh-fragment.glsl
```diff
@@ -102,6 +107,34 @@ void main() {
     // Use specific color provided in color attribute. For example plate boundary color.
     // Note that rendering only for alpha > X (e.g. 0.75 or 0.5) makes this color line thinner.
     diffuseColor *= vec4(vColor.r, vColor.g, vColor.b, 1.0);
+  } else if (usePatterns == true) {
+    // This doesn't look convincing, but unfortunatelly that's the only way to do it in GLSL (=> version supported by
+    // WebGL to be more speciifc). It's impossible to simply say:
```
kswenson, line 112:
```suggestion
    // WebGL to be more specific). It's impossible to simply say:
```

js/plates-view/plate-mesh-fragment.glsl
```diff
@@ -102,6 +107,34 @@ void main() {
...
+      diffuseColor *= texture2D(patterns[6], vUv * patternScale[6]);
+    } else if (vPatternIdx == 7) {
+      diffuseColor *= texture2D(patterns[7], vUv * patternScale[7]);
+    } else if (vPatternIdx == 8) {
+      diffuseColor *= texture2D(patterns[8], vUv * patternScale[8]);
+    } else if (vPatternIdx == 9) {
+      diffuseColor *= texture2D(patterns[9], vUv * patternScale[9]);
+    } else if (vPatternIdx == 10) {
+      diffuseColor *= texture2D(patterns[10], vUv * patternScale[10]);
+    }
```
kswenson, line 137: Not sure whether it's an improvement, but it seems like you could also use a switch statement:
```
  switch (vPatternIdx) {
    case 0: diffuseColor *= texture2D(patterns[0], vUv * patternScale[0]); break;
    ...
    case 10: diffuseColor *= texture2D(patterns[10], vUv * patternScale[10]); break;
  }

kswenson (approved):
👍 Looks good -- just a few minor comments/suggestions. I'm not familiar with WebGL programming, so I can't speak to that particularly.

js/plates-view/plate-mesh-fragment.glsl
pjanik: Right, it's more concise at least. I've just refactored this fragment.

PR #134, opened 2021-11-05
Remove rock colors, update key to handle colormap == rock

pjanik:
[#179943345]

Now, when we render rock patterns in the 3d view, I think it doesn't make sense to keep the hex color of the rocks. It's additional maintenance and they're already out of sync with the patterns. 

Also, I've updated the key to show rock key when colormap == rock.


kswenson (approved):
👍 Code changes look fine -- given the limitations of the patterning, are we sure that it's not premature to remove the coloring as an option?

pjanik:
> 👍 Code changes look fine -- given the limitations of the patterning, are we sure that it's not premature to remove the coloring as an option?

Hm, true. But Amy has seen the current implementation and she's fine with it. The rendering code is commented out, as it's more complex. I mostly removed bits that should be easy to restore, but annoying to maintain.

PR #135, opened 2021-11-05
Fix a bug that was sometimes leaving the boundary highlighted forever

pjanik:
[#179577787]

This was happening when the user wasn't closing the boundary type dialog, but moving his mouse straight to "Next" button.


kswenson (approved):
👍 LGTM

PR #136, opened 2021-11-06
VCR buttons (play, pause, step, restart)

kswenson:
[[#178645134]](https://www.pivotaltracker.com/story/show/178645134) Step, Start, Pause Buttons
[[#180019386]](https://www.pivotaltracker.com/story/show/180019386) Restart Button

Demo: https://tectonic-explorer.concord.org/branch/vcr-buttons/index.html?preset=benchmark

@pjanik I created a `bottom-bar` branch which this PR (branch `vcr-buttons`) is based on so that we can merge multiple PRs into the `bottom-bar` branch until we get the go-ahead to merge `bottom-bar` into `master`.

After looking at the implementations in `flooding-model` and `dese-models` (which is a slightly newer version of the code in `@concord-consortium/react-components`), I decided to start from the `dese-models` versions. The main reasons were that the code is more recent and presumably reflects more recent design thinking and because the "bubbles" implementation seems cleaner. We had to move away from the border-covers model in DESE because we kept running into circumstances in which the covers would fail to hide the borders under certain combinations of window size and browser zoom. The DESE version does bring theming support via CSS custom properties, which we don't intend to use any time soon, but I don't think it's a bad thing to have those properties defined even without taking advantage of theming. They provide a certain amount of self-documentation and if we ever do decide to support a dark mode or some other form of theming we'll have a leg up.

For this PR I focused mainly on the VCR buttons themselves and not on the styling of the bottom bar itself.


pjanik (approved):
Looks good to me, thanks.
I think one day we could convert `.less` to `.scss` as I think this standard has won within CC projects. And decide whether to fully go with css modules or the opposite. But none of these things is related to this PR, just random thought when I was looking at lots of styles. :wink:

kswenson:
>I think one day we could convert .less to .scss as I think this standard has won within CC projects.

I thought about that as I was working on this PR because, of course, I had to convert them from `.scss` to `.less`, but since all that was required was changing `$` to `@` in a few places, doing so was the path of least resistance.

>And decide whether to fully go with css modules or the opposite.

In this case it was nice that the Webpack config supports a naming convention to opt in/out of CSS modules so I didn't have to convert these files to CSS modules as part of the migration.

PR #137, opened 2021-11-08
Fix bugs found during QA

pjanik:
This PR fixes two bugs mentioned here:
https://www.pivotaltracker.com/story/show/180088420/comments/227868961

Force arrow fix isn't perfect, but it just works okay in practice. I don't think it's worth trying to calculate real intersections and move them up only when they're colliding with something.

### DIFF
--- a/js/components/planet-wizard.tsx
+++ b/js/components/planet-wizard.tsx
@@ -175,6 +175,7 @@ class PlanetWizard extends BaseComponent<IProps, IState> {
 
   setForcesStep() {
     const { setOption } = this.simulationStore;
+    this.simulationStore.setAnyBoundaryDefinedByUser(false);
     setOption("interaction", "assignBoundary");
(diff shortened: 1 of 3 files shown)


kswenson (approved):
👍 Looks good. One minor nit -- you might consider calling `setAnyBoundaryDefinedByUser()` in `setSelectedBoundaryType()` (rather than setting `anyBoundaryDefinedByUser` directly) for the benefit of anyone searching for `setAnyBoundaryDefinedByUser()` down the road to determine where it gets set.

pjanik:
> 👍 Looks good. One minor nit -- you might consider calling `setAnyBoundaryDefinedByUser()` in `setSelectedBoundaryType()` (rather than setting `anyBoundaryDefinedByUser` directly) for the benefit of anyone searching for `setAnyBoundaryDefinedByUser()` down the road to determine where it gets set.

I just did it in the last commit, thanks.

PR #139, opened 2021-11-08
Magma that makes it all the way to the surface should solidify

pjanik:
https://www.pivotaltracker.com/story/show/179940138

@kswenson, I've tried to explain updates in person. The model should make more sense now and be a bit simpler. I had to update Cypress tests that were checking the properties of fields with volcanoes.

The model should produce similar volcanic mountains chains. The new version is on the left side, the old one is on the right:
Screenshot 2021-11-08 at 18 53 40

### DIFF
--- a/js/plates-model/volcanic-activity.ts
+++ b/js/plates-model/volcanic-activity.ts
@@ -8,7 +8,7 @@ export interface IMagmaBlob {
   dist: number;
   maxDist: number;
   xOffset: number;
-  canErupt: boolean;
+  isErupting: boolean;
   finalRockType: Rock | undefined;
 }
 
@@ -17,7 +17,6 @@ export interface ISerializedVolcanicAct {
   highVolcanoCapacity: number;
   magma: IMagmaBlob[];
   eruptionTime: number;
-  eruptionCooldown: number;
   // .intensity and .colliding are dynamically calculated every simulation step.
 }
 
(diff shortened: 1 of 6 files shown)


js/plates-model/volcanic-activity.ts
```diff
@@ -171,29 +162,30 @@ export default class VolcanicActivity {
...
-        this.eruptionTime = 0;
-        this.eruptionCooldown = ERUPTION_COOLDOWN;
-      }
+    }
+    if (this.eruptionTime <= 0 || !this.active) {
+      this.eruptionTime = 0;
+      this.magma.forEach(blob => {
+        if (blob.isErupting) {
+          blob.isErupting = false;
+        }
```
kswenson, line 187: Is this functionally different than the simpler:
```suggestion
        blob.isErupting = false;
```
I thought MobX was smart enough to not notify observers if the value didn't actually change. Anyway, more a question than a recommendation.

kswenson (approved):
👍 LGTM

js/plates-model/volcanic-activity.ts
pjanik: Actually, there's no MobX anywhere in the plates-model dir. It's all model files used by the web worker. I guess I've added this `if` to limit the number of writes (as theoretically should be slower than value comparison 😉).  Usually, this would deserve to be called be premature optimization, but maybe it won't hurt in files in `plates-model` dir.

PR #141, opened 2021-11-08
Don't render plate velocity when rock patterns are shown

pjanik:
[#179943345]

This PR fixes a bug / missed UI specs: https://www.pivotaltracker.com/story/show/179943345/comments/227876972


kswenson (approved):
👍 LGTM

PR #143, opened 2021-11-09
Bottom bar: draw cross-section and take rock sample buttons

kswenson:
[[#180018736]](https://www.pivotaltracker.com/story/show/180018736)

Demo: https://tectonic-explorer.concord.org/branch/icon-buttons/index.html?preset=benchmark

Implements the `Draw Cross-section` and `Take [Rock] Sample` buttons using a shared `<IconHighlightButton>` component.

@pjanik Note that I left some of the old code in place but commented out because I'm not sure what might be needed for Geode. I'm happy to actually remove any such code that you don't think we'll need.


pjanik (approved):
Looks good to me.

Re old code - even though I designed it personally, I think it's fine to remove it  💔 😉 . I don't see us going back to this version. The only place where we might need the interactions selector is PlanetWizard. If I remember correctly, we don't have new UI specs for that yet.

PR #144, opened 2021-11-09
Fix bugs related to interactions and cursors

pjanik:
https://www.pivotaltracker.com/story/show/180230314

There are two instances of interaction managers - one for the planet view and one for the cross-section. Recently, I've moved event handlers from canvases to document. This was breaking interactions in some edge cases, depending on the timing of MobX observers. For example, the cross-section manager could remove all the event handlers from document **after** the planet view manager just added its own handlers.

Also, I've fixed cursors handling. Using document was causing similar issues, so I've limited custom cursor to canvas that is unique per manager.

### DIFF
--- a/js/plates-interactions/base-interactions-manager.ts
+++ b/js/plates-interactions/base-interactions-manager.ts
@@ -3,7 +3,7 @@ import $ from "jquery";
 import { EventEmitter2 } from "eventemitter2";
 import { IInteractionHandler, mousePos, mousePosNormalized } from "./helpers";
 
-const NAMESPACE = "interactions-manager";
+let _instanceId = 0;
 
 export interface IView {
   domElement: HTMLElement;
@@ -14,6 +14,7 @@ export interface IView {
 }
 
 export class BaseInteractionsManager {
+  namespace: string;
   activeInteraction: IInteractionHandler | null;
   emitter: EventEmitter2;
   interactions: Record<string, IInteractionHandler> = {};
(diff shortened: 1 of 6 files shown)


js/plates-interactions/base-interactions-manager.ts
```diff
@@ -34,14 +37,20 @@ export class BaseInteractionsManager {
 
   setInteraction(name: string) {
     if (this.activeInteraction) {
-      this.activeInteraction.setInactive();
+      if (this.activeInteraction.cursor) {
+        this.view.domElement.style.cursor = "auto";
+        console.log("set cursor auto");
```
kswenson, line 42:
```suggestion
```
Do you want to leave this in?

js/plates-interactions/base-interactions-manager.ts
```diff
@@ -34,14 +37,20 @@ export class BaseInteractionsManager {
...
+      }
       this.activeInteraction = null;
       this.disableEventHandlers();
     }
     if (name && this.interactions[name]) {
       this.activeInteraction = this.interactions[name];
-      this.activeInteraction.setActive();
       this.enableEventHandlers();
+      if (this.activeInteraction.cursor) {
+        console.log("set cursor", this.activeInteraction.cursor);
```
kswenson, line 51:
```suggestion
```
Do you want to leave this in?

kswenson (approved):
👍 Looks good -- just a question about the console.logs.

js/plates-interactions/base-interactions-manager.ts
pjanik: No, thanks. :)

PR #146, opened 2021-11-10
Bottom bar styling

kswenson:
PT: [[#180169391]](https://www.pivotaltracker.com/story/show/180169391) (in part)

Demo: https://tectonic-explorer.concord.org/branch/bottom-bar-styling/index.html?preset=benchmark

Adjust bottom bar flex containers so that bubbles rise above bar

### DIFF
--- a/css-modules/sidebar-menu.less
+++ b/css-modules/sidebar-menu.less
@@ -1,7 +1,7 @@
 .sidebar {
   min-width: 300px;
   padding-top: 15px;
-  height: calc(~"100% - 56px");
+  height: calc(~"100% - 65px");
 
   .buttonContainer {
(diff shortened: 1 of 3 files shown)


pjanik:
Looks good. I can see one glitch - when I open the sidebar (that we might still have for a while), the bottom bar gets squeezed in and some weird things happen to the menu and full-screen button. But I guess this can be fixed in another PR too, up to you.

pjanik (approved):

kswenson:
>I can see one glitch - when I open the sidebar (that we might still have for a while), the bottom bar gets squeezed in and some weird things happen to the menu and full-screen button.

Should be fixed now.

PR #147, opened 2021-11-11
Add earthquakes/volcanic eruption switches to bottom bar

kswenson:
PT: [[#180019424]](https://www.pivotaltracker.com/story/show/180019424)

Demo: https://tectonic-explorer.concord.org/branch/volcanoes-earthquakes/index.html?preset=benchmark

Presence/absence of switches is controlled by authoring (same as before)
Clicking on any part of switch changes the polarity of the switch
Thumb can also be dragged from one end to the other

### DIFF
--- a/css-modules/slider-switch.less
+++ b/css-modules/slider-switch.less
@@ -1,0 +1,70 @@
+@import "../css/variables.less";
+
+.sliderSwitch {
+  position: relative;
+  display: flex;
+  align-items: center;
+
+  &:hover {
+    .thumbHighlight {
+      background: #ffffff80;
+    }
+  }
+  &:active {
+    .thumbHighlight {
+      background: #ffffff;
+    }
+  }
+
+  .track {
+    width: 27px;
+    height: 10px;
+    margin-left: 20px;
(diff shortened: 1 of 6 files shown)


pjanik (approved):
It looks good. 👍 

The draggable switch is pretty fancy. Have you seen that? :wink:
https://mui.com/components/switches/
I've implemented pretty much identical switches (Michael uses them across projects) using this component. I think in Wildfires and/or Flood model, but over time, they got replaced with regular sliders for some reason. Or did you decide to use the custom component because of the dragging?

I don't think it's worth refactoring back to this Material UI component, yours is still pretty compact and it works fine (or even better with this dragging abilities).

The only issue I've found is that it doesn't seem to work with touch events. I mean you can drag, but tapping doesn't change its state (and I bet most users will click/tap rather than drag).

I'll merge this PR and the touch event support can be fixed separately by you or me later.

pjanik:
Also, I think we'll have to refactor `sidebar` URL config to something else, as it's starting to make less sense now. It's nice it's backward compatible though. But it also seems like a separate story.

PR #149, opened 2021-11-13
Fix touch switches

kswenson:
Fixes behavior of `<SliderSwitch>` component (used for the `Volcanoes` and `Earthquakes` switches on the bottom bar) on touch devices pointed out in review of #147.

Demo: https://tectonic-explorer.concord.org/branch/fix-touch-switches/index.html?preset=benchmark

### DIFF
--- a/js/components/slider-switch.tsx
+++ b/js/components/slider-switch.tsx
@@ -11,8 +11,10 @@ interface IProps {
 export const SliderSwitch: React.FC<IProps> = ({ label, isOn, onSet }) => {
   const kThumbOffPos = 0;
   const kThumbOnPos = 20;
+  const kClickTime = 250;
   const [isDragging, setIsDragging] = useState(false);
   const totalDrag = useRef(0);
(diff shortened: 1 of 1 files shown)


pjanik (approved):
Looks good, thanks.

Sometimes I can see following error in the console:
Screenshot 2021-11-15 at 13 52 37
I was using Chrome in the touch-device mode, and just tapping the toggle quite a few times. But it doesn't seem to break anything, so it might not be worth investigating.

kswenson:
> Looks good, thanks.
> 
> Sometimes I can see following error in the console: Screenshot 2021-11-15 at 13 52 37 I was using Chrome in the touch-device mode, and just tapping the toggle quite a few times. But it doesn't seem to break anything, so it might not be worth investigating.

Looks internal to react-draggable: https://github.com/react-grid-layout/react-draggable/issues/553. I vote we ignore it for now and hope they fix it in a future update to the library.

PR #151, opened 2021-11-15
Tabs

pjanik:
This PR is ready to be reviewed, although is **not** ready to be merged into master.

I'll keep this PR open for a while, merging subsequent PRs into `179940030-tabs` branch. Pretty much following the bottom-bar branch pattern.

I think it's still worth reviewing the current, incomplete state - this PR implements functional and configurable tabs. And the diff should be much more readable than one including all the cutting and refactoring of the `color-key.tsx` component - lots of code will be moved around.

Tab names follow the latest recommendations from Amy and Michael. I've also made everything configurable using the authoring page. Unrelated bonus: I've restyled autocomplete component, as it was really hard for me to read it and see where's the boundary of this component, and where the next option starts. I think making them wider and adding borders cleans them up a lot.

Demo (when the deployment is done):
https://tectonic-explorer.concord.org/branch/tabs/index.html?preset=subduction


css-modules/side-container.less
```diff
@@ -0,0 +1,79 @@
...
+  border-top-right-radius: 9px;
+  border-width: 4px;
+  border-bottom: none;
+  background: white;
+  margin-right: 6px;
+  //box-shadow: 0 0 0 1px #797979;
+  vertical-align: bottom;
+  color: #434343;
+  padding: 0;
+  font-size: 14px;
```
kswenson, line 34: Font-family for tab labels should be Lato and font-weight should be bold.

js/stores/simulation-store.ts
```diff
@@ -453,6 +462,14 @@ export class SimulationStore {
     this.anyHotSpotDefinedByUser = val;
   }
 
+  @action.bound setKeyActive(val: boolean) {
```
kswenson, line 465: "Active" makes me think of the CSS ":active" state, i.e. selected. Perhaps something like `setKeyVisible`, `setKeyVisibility`, or `setKeyShowing`?

css-modules/side-container.less
```diff
@@ -0,0 +1,79 @@
...
+.tab {
+  cursor: pointer;
+  display: inline-block;
+  position: relative;
+  list-style: none;
+  margin-bottom: -1px;
+
+  border-style: solid;
+  border-top-left-radius: 9px;
+  border-top-right-radius: 9px;
```
kswenson, line 25: I would generally write this as
```suggestion
  border-radius: 9px 9px 0 0;
```

kswenson (approved):
👍 The formatting of the autocomplete controls on the authoring page is a nice improvement. The initial tab code looks fine for what it does so far. I pointed out below that the tab labels don't currently have the specced family and weight. Otherwise, this seems like a fine start to the tabs work.

css-modules/side-container.less
pjanik: Thanks, I've updated that. Actually, I've just globally changed the default font from Museo to Lato. I think Museo was my old choice (based on Material UI?). Michael seems to use Lato for all the components.

js/stores/simulation-store.ts
pjanik: True, I've changed that to `setKeyVisible`.

css-modules/side-container.less
pjanik: Updated 👍

pjanik:
I've fixed all the issues and merged `master` into this branch resolving some conflicts.
I'm assuming that at this point this PR is reviewed and approved. 
I'll keep merging subsequent PRs into this branch, and once things are functional, this PR will get merged into `master`.

PR #152, opened 2021-11-15
Eliminate metamorphism option from authoring and sidebar in geode

kswenson:
From a comment in [this PT story](https://www.pivotaltracker.com/story/show/180168841):

>I also noticed that the "Metamorphism" check box was available in the menu in the link provided below. That should not be in the GEODE version. Can you remove it?

Testable at https://tectonic-explorer.concord.org/branch/no-geode-metaphorphism/index.html?geode&preset=benchmark.


pjanik (approved):
Looks good 👍

PR #153, opened 2021-11-15
Dependency updates

kswenson:
Dependencies _not_ updated:

#### ESM-only modules
- `array-move`, `d3-color`, `d3-interpolate`, `d3-scale`, `screenfull`

#### firebase
- v9 introduces a completely new API

#### three
- updating to 0.131 rendered the model but broke cypress tests
- updating to 0.132+ failed to render the model with shader errors in the console


pjanik (approved):
Looks good, thanks!

PR #154, opened 2021-11-16
Fix fullscreen icon button

kswenson:
For as-yet-unknown reasons, the update to `css-loader` v6 broke the rendering of the fullscreen icon button. In lieu of investigating we simply revert to `css-loader` v5 for now.

Demo: https://tectonic-explorer.concord.org/branch/css-loader-5/index.html?preset=benchmark


pjanik (approved):
👍

PR #157, opened 2021-11-18
Add GEODE rock key, refactor color definitions, hide interaction buttons in TecRock, fix rock key visibility bug

pjanik:
https://www.pivotaltracker.com/story/show/180330671

This PR includes three semi-related updates - commit names should describe them well. 
They're all related to TecRock and GEODE modes.
Note that GEODE mode is activated by adding ?geode URL param.


knowuh (approved):
👍

PR #162, opened 2022-01-03
Tectonic Explorer Automation

saravanan-udayakumar:


pjanik:
It seems they didn't work on GH Actions - the Cypress tests were canceled after 15mins.

saravanan-udayakumar:
@pjanik , this PR is ready for review. We removed the test authoringcontrol-maptype.js, which is causing the problem. Will fix those test in separate branch.

pjanik (approved):
Thanks for investigating this. Looks good, and it seems to work fine now in the GH Actions env.

tejal-shah (approved):

PR #163, opened 2022-01-25
Add LARA / AP logging

pjanik:
[#178132549]

I've wrapped the lara-interactive-api `log` function in a custom wrapper defined at `js/log.ts`. The main goal of that was to enforce self-documentation using types. It won't be possible to add a new event without documenting it using a new type. I'm planning to provide a link to `js/log.ts` to the project team, so they can always check all the available log messages. I'm trying to keep this file minimal and readable for non-programmers. If I use some external type, I'm adding a comment that lists possible values.

We also have a google spreadsheet with log events that I can't find now. But once I do, I think I'll just copy there a link to this file too. I think this is a more maintainable approach long-term than keeping a separate spreadsheet list.

### DIFF
--- a/js/components/bottom-panel.tsx
+++ b/js/components/bottom-panel.tsx
@@ -15,6 +15,7 @@ import { IGlobeInteractionName } from "../plates-interactions/globe-interactions
 import { BaseComponent, IBaseProps } from "./base";
 
 import "../../css/bottom-panel.less";
+import { log } from "../log";
 
 function toggleFullscreen() {
   if (screenfull.isEnabled) {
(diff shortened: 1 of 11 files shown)


js/components/bottom-panel.tsx
```diff
@@ -15,6 +15,7 @@ import { IGlobeInteractionName } from "../plates-interactions/globe-interactions
 import { BaseComponent, IBaseProps } from "./base";
 
 import "../../css/bottom-panel.less";
+import { log } from "../log";
```
kswenson, line 18: Import order

kswenson (approved):
👍 Looks good!

PR #164, opened 2022-01-25
Plate draw tool: make continent edges softer

pjanik:
[#179940164

Old version of continent drawing:
old

A new one, implemented in this PR:
new

I've updated the distance function to use real distance instead of the number of fields between given field and the center.


kswenson (approved):
👍 LGTM

PR #168, opened 2022-01-27
Rock key: scroll selected rock into view

pjanik:
[#180480249]

I've added 250ms delay to flash animation so there's time for the container to scroll. Sometimes it's necessary, sometimes not, but 250ms seems to be a good compromise. As even if scrolling isn't necessary, it doesn't feel too strange.


kswenson (approved):
👍 LGTM

PR #170, opened 2022-01-27
Planet Wizard: keep user changes while he navigates between steps

pjanik:
[[#180479948]](https://www.pivotaltracker.com/story/show/180479948)

I was going back and forth here, but I think the cleanest solution to address request in this PT story is just to never reset/unload user changes. Unless he goes to the very initial step (layout selection). That makes implementation and behavior clean.


kswenson (approved):
👍 Looks good -- implement a new feature by deleting code -- I like it! Now if you could just figure out how to generalize the pattern so that all features can be implemented by deleting code. 🤔 😉

PR #171, opened 2022-01-28
Fix smoke test

pjanik:
Well, it took a few attempts, but I think failures were caused by the fact that now the reload button has a small delay before page refresh is executed (logging). So, we're triggering the click, but sometimes cypress was starting to check the existing DOM too soon (before reload happened).


kswenson (approved):
👍 LGTM

PR #172, opened 2022-01-28
Update dependencies, update code to use Firebase API v9

pjanik:
I've updated most of the dependencies. I've used SVGR settings from starter-project, as they seem to make more sense than defaults. Also, Firebase v9 changes API completely, so I had to update storage.ts file.


js/storage.ts
```diff
@@ -15,23 +17,32 @@ export function initDatabase() {
...
+  if (!app) {
+    return;
+  }
+  const db = getDatabase(app);
   const uuid = uuidv4();
 
-  db.ref("models/" + uuid).set({
+  const dbRef = ref(db, "models/" + uuid);
+  set(dbRef, {
     // JSON.strignify + JSON.parse will remove all the undefined values from the object.
```
kswenson, line 28:
```suggestion
    // JSON.stringify + JSON.parse will remove all the undefined values from the object.
```

webpack.config.js
```diff
@@ -109,10 +109,30 @@ module.exports = {
...
+                        // https://github.com/svg/svgo/blob/master/plugins/convertShapeToPath.js
+                        convertShapeToPath: false,
+                        // leave "class"es and "id"s alone
+                        // https://github.com/svg/svgo/blob/master/plugins/prefixIds.js
+                        prefixIds: false,
+                        // leave "stroke"s and "fill"s alone
+                        // https://github.com/svg/svgo/blob/master/plugins/removeUnknownsAndDefaults.js
+                        removeUnknownsAndDefaults: { defaultAttrs: false },
+                        // leave viewBox alone
+                        removeViewBox: false
```
kswenson, line 131: If it's really true that svgo removes the `viewBox` by default, then we should probably add this to the default configuration for CLUE, DESE, starter-projects, etc. Odd that I've never noticed that. 🤔

js/storage.ts
```diff
@@ -1,10 +1,12 @@
-import firebase from "firebase/app";
-import "firebase/database";
-import { v4 as uuidv4 } from "uuid";
+import { FirebaseApp, initializeApp } from "firebase/app";
+import { set, ref, onValue, getDatabase } from "firebase/database";
```
kswenson, line 2: I'm glad to see that you took the plunge to the new v9 API without too much difficulty. I've been holding off on upgrading on several projects that use firebase more extensively. From this example, however, it looks like the changes are pretty mechanical.

kswenson (approved):
👍 Looks good! A few inline comments below.

webpack.config.js
pjanik: The main splash screen in TE is broken without that rule:
Screenshot 2022-01-27 at 17 48 58
It was also visible when I inspected this SVG and tried to re-add viewBox by hand.

js/storage.ts
pjanik: They are, but almost every line will be affected. 😉 
Some functions are gone too and there's no direct equivalent (like `once` => `onValue` + custom options). But the new API definitely looks nicer, less magic (especially with the app config), more explicit.

kswenson: >The splash screen in TE is broken without that rule

That also suggests that it's a new default optimization in v6 given that you weren't experiencing it before. Given that, I'm surprised we haven't seen more breakage. 🤔

webpack.config.js
kswenson: >The splash screen in TE is broken without that rule

That also suggests that it's a new default optimization in v6 given that you weren't experiencing it before. Given that, I'm surprised we haven't seen more breakage. 🤔

pjanik: Actually, it was already in v5 - you can take a look at the diff here. The old SVGR version was v5, and the old config was:
```
plugins: {
  // SVGR removes viewbox by default. Disable this behavior.
  removeViewBox: false
}
```
I looked at the starter-projects, as I couldn't disable `removeViewBox` using weird SVGR v6 options / plugins syntax. :wink: And then transferred other options too.

Re viewBox itself - maybe we're just lucky. Most of the SVG files should work fine with viewBox. Even when I look at splash screen here, I don't understand why it needs viewBox to work fine (but I confirmed that using devtools). Or... maybe SVGr does some magic transformations of all the SVG element coordinates so it can remove viewBox (I wouldn't be surprised look how many attributes it modifies).

kswenson: Makes sense -- all the more reason to make it part of our default configuration. As far as I'm concerned we should disable any optimization that can potentially break things.

PR #176, opened 2022-02-03
Reset all the options on model reload + hide some UI elements in Planet Wizard

pjanik:
This PR fixes issue pointed out by Saravanan (see the original [comment](https://www.pivotaltracker.com/story/show/180480267/comments/229323808)).

He noticed that the earthquakes caveat is being displayed in the Planet Wizard. This should not happen anymore. I also realized we didn't reset most of the options on model reload. So, the user could go to advanced options, turn on things like wireframe rendering, and he'd be stuck with it in the Planet Wizard mode after reload (aka "plate reset"). There's no clear answer if we should reset all the UI options or not, but it definitely feels safer, as it's less likely users will be stuck with a config that isn't working well in Planet Wizard.

### DIFF
--- a/js/components/planet-view.tsx
+++ b/js/components/planet-view.tsx
@@ -120,12 +120,12 @@ export default class PlanetView extends BaseComponent<IBaseProps, IState> {
   }
 
   render() {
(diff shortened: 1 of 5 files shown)


kswenson (approved):
👍 LGTM

PR #177, opened 2022-02-04
Add continental shelf around plate boundary

pjanik:
This PR should fix an issue described here:
https://www.pivotaltracker.com/story/show/181127412/comments/229373533

Previously, the plate boundary was preventing the draw tool from adding a smooth slope / continental shelf. Now, this should be fixed.

We could try to do it a bit faster than running a second BFS for each field. But since we're dealing with just a few fields, I think this shouldn't matter too much in practice.


kswenson (approved):
👍 Looks good!

PR #180, opened 2022-02-17
Revert "Add continental shelf around plate boundary"

pjanik:
Reverts concord-consortium/tectonic-explorer#177

Finally, Amy didn't like the continental shelf being automatically added around boundaries, so I'm revering the PR that added that. See:
https://www.pivotaltracker.com/story/show/181127412
https://www.pivotaltracker.com/story/show/181127412/comments/229679395


kswenson (approved):
👍 And the implementation was so elegant!

PR #182, opened 2022-02-20
Make user-defined forces stronger to enhance continental collision

pjanik:
[#181153609]

This PR makes user-defined forces (ones specified in Planet Wizard) much stronger and configurable. Previous values were too low and the orogeny wasn't really happening.


kswenson (approved):
👍 Looks good!

PR #183, opened 2022-02-20
Fix issues in Crust class that could make some rock layers have negative thickness

pjanik:
[#181320915

This PR fixes issue shown below. These "lakes" behind the mountains were caused by rock layers that had negative thickness. It was really confusing, as the problem wasn't visible in the cross-section and only in the 3D view. First, I thought it was caused by the erosion code, but it was actually `subductOrFold` method. 

I've added `Math.min(1, ...)` each time where rock layer thickness was reduced. Also, I've added a simple test in functions that are most likely to detect this negative value.

https://user-images.githubusercontent.com/767857/154832794-83f1e377-34ec-4775-a929-aaa6bb2b62fd.mov

issue 1
]

### DIFF
--- a/js/plates-model/crust.ts
+++ b/js/plates-model/crust.ts
@@ -4,8 +4,8 @@ import { FieldType } from "./field";
 import getGrid from "./grid";
 import { Rock, rockProps } from "./rock-properties";
 
-export interface IRockLayer { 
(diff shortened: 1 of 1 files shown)


kswenson (approved):
👍 Looks good!

PR #184, opened 2022-02-23
Update plate forces, skip some of the Cypress tests

pjanik:
https://www.pivotaltracker.com/story/show/181153609

Small updates related to plate forces. I've increased max plate velocity so the new param works with higher values too. Also, I generally increased forces but also made force decrease faster. This should accelerate plates faster, so the results are less dependent on the distance between continents.

This last update broke some model Cypress tests. They're very specific and bound to the current model behavior. They'll be way more changes to model behavior, so I think it makes sense to skip these problematic tests for now and update them at the end of dev cycle.


kswenson (approved):
👍 LGTM

PR #190, opened 2022-02-27
181392960 time slider

pjanik:
https://www.pivotaltracker.com/story/show/181392960

Time slider used to change timestep. I've been trying to include timestep in all the simulation rules, but it's still difficult to ensure that model produces a very similar result for various time step values. It is not well-defined and well-known math that could be tested and verified.

So, I've decided to change the time slider so it affects the number of calculations per second. This is a safer approach, even though it might be limited by the performance of the machine. But project team was more concerned about the model being too fast, not too slow, and slowing down should always work.

### DIFF
--- a/js/components/advanced-options.tsx
+++ b/js/components/advanced-options.tsx
@@ -31,7 +31,7 @@ interface IState {}
 @observer
 export class AdvancedOptions extends BaseComponent<IProps, IState> {
   changeInteraction: (value: any) => void;
-  changeTargetModelStepsPerSecond: (value: any) => void;
+  changeTargetModelStepsPerSecond: (value: number) => void;
   toggleBoundaries: () => void;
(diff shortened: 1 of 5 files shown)


js/components/advanced-options.tsx
```diff
@@ -31,7 +31,7 @@ interface IState {}
 @observer
 export class AdvancedOptions extends BaseComponent<IProps, IState> {
   changeInteraction: (value: any) => void;
-  changeTimestep: (value: any) => void;
+  changeTargetModelStepsPerSecond: (value: any) => void;
```
kswenson, line 34:
```suggestion
  changeTargetModelStepsPerSecond: (value: number) => void;
```
I would think this could be typed at this point.

kswenson (approved):
👍 LGTM

PR #197, opened 2022-03-03
Update tabs UI to match specs

pjanik:
https://www.pivotaltracker.com/story/show/181127423

This PR adds hover and active states to tabs + updates their widths.


kswenson (approved):
👍 LGTM

PR #198, opened 2022-03-04
Update Rock Key to differentiate different kinds of metamorphic rocks in each grade

pjanik:
https://www.pivotaltracker.com/story/show/180970003

This PR makes a few significant changes to Rock Key. The project team wanted to show different diagrams for Low/Medium/High Grade Metamorphic Rocks depending on whether the user clicked the rock in subduction zone or continent-continent collision.

However, they still wanted to show only one header for each grade. So, TecRock will expand this header with a different diagram, depending on the picked rock. If the user clicks the header directly, it'll open the Continental Collision variant by default. I find this interaction pretty weird, but it was the requirement, and I didn't have a better idea how to handle that (apart from showing 7 different headers in the key, but it wouldn't look great).

That's why Rock Key has now a strange entry called `oneOf` (naming inspired by webpack configuration). It's just a group of rocks, displayed under one header in the key.

I've also added Contact Metarmopism as a separate entry. Users can click on the magma blob border to open it. That's why the border is now wider.


js/components/keys/rock-types.tsx
```diff
@@ -53,14 +57,16 @@ const FLASH_ANIMATION_DURATION = 750;
 interface IRockDef {
   name?: RockKeyLabel;
   shortName?: string; // if different than full name
-  pattern: string;
+  pattern?: string;
   image?: JSX.Element;
   diagram?: JSX.Element;
   notes?: JSX.Element;
+  oneOf?: Omit<IRockDef, "oneOf">[];
```
kswenson, line 64: Clever!

kswenson (approved):
👍 Looks good -- nicely done

PR #199, opened 2022-03-04
Make orogeny strength configurable

pjanik:
[#181456402]

### DIFF
--- a/js/plates-model/physics/forces.ts
+++ b/js/plates-model/physics/forces.ts
@@ -4,7 +4,7 @@ import Plate from "../plate";
 
 const BASIC_DRAG_FORCE_MOD = 0.000001 * config.friction;
 // Note that OROGENY_FORCE_MOD also includes config.friction.
-const OROGENY_FORCE_MOD = 0.000001 * config.friction / config.orogenyStrength;
+const OROGENY_FORCE_MOD = BASIC_DRAG_FORCE_MOD / config.orogenyStrength;
 
 // Basic drag force of the tectonic plate. It's very small, but it keeps model stable.
 export function basicDrag(field: Field) {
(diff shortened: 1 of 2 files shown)


js/plates-model/physics/forces.ts
```diff
@@ -2,11 +2,13 @@ import config from "../../config";
 import Field from "../field";
 import Plate from "../plate";
 
-const FORCE_MOD = 0.000001 * config.forceMod;
+const BASIC_DRAG_FORCE_MOD = 0.000001 * config.friction;
+// Note that OROGENY_FORCE_MOD also includes config.friction.
+const OROGENY_FORCE_MOD = 0.000001 * config.friction / config.orogenyStrength;
```
kswenson, line 7:
```suggestion
const OROGENY_FORCE_MOD = BASIC_DRAG_FORCE_MOD / config.orogenyStrength;
```
Seems like the comment would be reinforced and you wouldn't have to duplicate the constant factor.

Also, more for my understanding and to validate that the names make sense, does it make sense that as `orogenyStrength` increases, `OROGENY_FORCE_MOD` decreases?

kswenson (approved):
👍 LGTM -- just one minor suggestion and one question

js/plates-model/physics/forces.ts
pjanik: > Seems like the comment would be reinforced and you wouldn't have to duplicate the constant factor.

Done, thanks.

> Also, more for my understanding and to validate that the names make sense, does it make sense that as orogenyStrength increases, OROGENY_FORCE_MOD decreases?

Yes. Orogeny force is the friction force that is applied to two continents when they collide. So, when this force is smaller, two plates will overlap more and the resulting mountains are bigger. So, orogeny is "stronger". I used that name so Amy / the project team doesn't have to think about forces. Just size / intensity of the orogeny process.

PR #200, opened 2022-03-08
Add Planet Wizard validation pattern and check user-set densities

pjanik:
https://www.pivotaltracker.com/story/show/181257654

A quite common problem in TecRock is that users draw a single continent at the boundary, set boundary type to be convergent, but they set the continental plate density to be higher than the neighboring oceanic plate. This results in an unallowed collision type that adds plenty of friction forces to stop continental subduction. So, the models comes to halt very quickly. Amy was concerned that users are getting confused by that.

This PR adds a new mechanism in Planet Wizard - each step might be validated and the user might be asked to make some changes. It's been used in the density selection step. So, if the validation function detects the use case described above, it'll suggest changing the density setup.

Demo: https://tectonic-explorer.concord.org/branch/density-dialog/index.html?planetWizard=true

A few cases are described below:

1. This setup should result in the warning.
Screenshot 2022-03-07 at 19 27 13

2. This setup should NOT result in the warning. There are some areas of the continent that will subduct, but I'm assuming this is a valid continent-continent collision.
Screenshot 2022-03-07 at 19 28 28

3. This setup will result in a warning no matter how the user setups densities. It cannot be fixed. But user can still select  "Continue anyway" and run the model to see what happens. Or go back to the continent drawing step.
Screenshot 2022-03-07 at 19 30 36

### DIFF
--- a/js/components/planet-wizard.tsx
+++ b/js/components/planet-wizard.tsx
@@ -11,6 +11,8 @@ import ccLogoSmall from "../../images/cc-logo-small.png";
 import SortableDensities from "./sortable-densities";
 import { BaseComponent, IBaseProps } from "./base";
 import { log } from "../log";
+import { isDensityDefinedCorrectly } from "../stores/helpers/planet-wizard-model-validators";
+import { Dialog, DialogActions, DialogContent, DialogContentText, DialogTitle } from "@mui/material";
 
 import "../../css/planet-wizard.less";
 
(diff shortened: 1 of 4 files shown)


js/components/planet-wizard.tsx
```diff
@@ -26,7 +28,18 @@ interface IStepsData {
   info: (geode: boolean) => string; // label of bottom bar button
   navigationDisabled?: boolean; // whether next/back navigation should be disabled globally
   nextDisabled?: (simulationStore: SimulationStore) => boolean; // whether next navigation should be disabled conditionally
+  validation?: IValidation; // when defined, the test test will be ran before user can go to the next step
```
kswenson, line 31:
```suggestion
  validation?: IValidation; // when defined, the test will be run before user can go to the next step
```

js/components/planet-wizard.tsx
```diff
@@ -26,7 +28,18 @@ interface IStepsData {
...
 }
+
+interface IValidation {
+  // If this function returns false, user will see a warning dialog that lets him return and fix the configuration.
```
kswenson, line 35:
```suggestion
  // If this function returns false, user will see a warning dialog that lets them return and fix the configuration.
```
An opportunity to use an inclusive pronoun.

kswenson (approved):
👍 Looks good

PR #201, opened 2022-03-09
Update/fix tabs layout

pjanik:
This PR fixes a few bugs found by Saravanan during QA:
https://www.pivotaltracker.com/story/show/181127423/comments/229981270
New variables should make tabs easier to work with.

### DIFF
--- a/css-modules/side-container.less
+++ b/css-modules/side-container.less
@@ -4,19 +4,21 @@
 @outlineColor: #797979;
 @outlineWithoutBottom: 0 -1px 0 @outlineColor, -1px 0 0 @outlineColor, 1px 0 0 @outlineColor;
 
+@tabHeight: 29px; // without border
+@tabSelectedHeight: 35px; // without border
+@tabTopPadding: 5px;
+
 @tabListMarginTop: 6px;
-@tabListHeight: 40px;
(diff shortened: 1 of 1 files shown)


css-modules/side-container.less
```diff
@@ -35,7 +37,7 @@
       display: inline-block;
       position: relative;
       list-style: none;
-      margin-bottom: -1px;
+      // margin-bottom: -1px;
```
kswenson, line 40: Do you want to keep this here or can it be removed now?

css-modules/side-container.less
```diff
@@ -61,11 +63,12 @@
 
     .tabSelected {
       .tabInsideContainer {
-        padding-top: 13px;
+        height: @tabSelectedHeight - @tabTopPadding;
+        // padding-top: 13px;
```
kswenson, line 67: Keep?

kswenson (approved):
👍 LGTM

css-modules/side-container.less
pjanik: Removed, thanks.

css-modules/side-container.less
pjanik: Removed, thanks.

PR #202, opened 2022-03-09
Update / cleanup presets

pjanik:
[#181411936]

I've removed some presets that didn't seem useful anymore and slightly updated `subduction` to have narrower continental shelf, as this preset is used quite often.


kswenson (approved):
👍

PR #206, opened 2022-04-21
Add temp/pressure measurement control to bottom bar

kswenson:
PT: [[#181680353]](https://www.pivotaltracker.com/story/show/181680353)
Demo: https://tectonic-explorer.concord.org/branch/tp-tool-control/index.html?preset=subduction

In the interest of dividing up the work into more manageable/reviewable chunks, we interpret this story to implement the `Measure Temp/Pressure` bottom bar control which can be used to enable/disable the `Measure Temp/Pressure` mode. While in this mode the cursor changes to a semitransparent circle when over the canvas that contains the cross-section view and to the browser's `not-allowed` cursor when over the globe or other non-measurable and non-clickable areas of the screen. The actual display of temperature/pressure alongside the cursor is left for the next story. The enabling/disabling of the `Measure Temp/Pressure` mode is also logged.

Not technically part of this story but fixed along the way -- appropriate log events are generated when disabling the `Draw Cross-section` and `Take Sample` modes. (Previously, the code logged `...Enabled` when enabling _or disabling_ the corresponding mode.)


js/plates-interactions/globe-interactions-manager.ts
```diff
@@ -27,6 +27,9 @@ export default class GlobeInteractionsManager extends BaseInteractionsManager {
       markField: new PlanetClick({ ...baseOptions, startEventName: "markField" }),
       continentDrawing: new PlanetClick({ ...baseOptions, startEventName: "continentDrawing", moveEventName: "continentDrawing", endEventName: "continentDrawingEnd" }),
       continentErasing: new PlanetClick({ ...baseOptions, startEventName: "continentErasing", moveEventName: "continentErasing", endEventName: "continentErasingEnd" }),
+      measureTempPressure: new PlanetClick({
+        ...baseOptions, cursor: "not-allowed", startEventName: "measureTempPressure",
+        moveEventName: "measureTempPressure", endEventName: "measureTempPressureEnd" }),
```
pjanik, line 32: Do we want to specify events here? It seems like they'll be emitted even though we don't support the temp-pressure tool in the globe view?

pjanik (changes requested):
Looks good. 👍 
I've just noticed one small issue - if you activate the temp-pressure tool and then close the cross-section, you're stuck in this mode, with `not-allowed` cursor, without the possibility to leave it (as the button is now disabled). This can be also handled in a subsequent PR if that's easier. Up to you.

js/plates-interactions/globe-interactions-manager.ts
kswenson: I was just following existing patterns. If they're not required then I'll remove them.

kswenson:
> Looks good. 👍 I've just noticed one small issue - if you activate the temp-pressure tool and then close the cross-section, you're stuck in this mode, with `not-allowed` cursor, without the possibility to leave it (as the button is now disabled). This can be also handled in a subsequent PR if that's easier. Up to you.

Oops -- seems worth fixing.

kswenson:
> > Looks good. 👍 I've just noticed one small issue - if you activate the temp-pressure tool and then close the cross-section, you're stuck in this mode, with `not-allowed` cursor, without the possibility to leave it (as the button is now disabled). This can be also handled in a subsequent PR if that's easier. Up to you.
> 
> Oops -- seems worth fixing.

Fixed in latest commit.

js/plates-interactions/globe-interactions-manager.ts
pjanik: That's the globe interactions manager, so I guess we won't have to deal with any events here, as this tool will not work with the globe view. I guess the only reason to keep this line is to set `cursor: "not-allowed` if we can't do it using plain CSS somewhere else.

pjanik (approved):
Looks good, thanks 👍

js/plates-interactions/globe-interactions-manager.ts
kswenson: CSS doesn't work because the cursor is set via inline style from the current interaction, so defining a globe interaction for setting the cursor seemed like the path of least resistance.

PR #207, opened 2022-04-22
Add temp/pressure measurement display [#180799854]

kswenson:
PT: [[#180799854]](https://www.pivotaltracker.com/story/show/180799854)
Demo: https://tectonic-explorer.concord.org/branch/tp-display/index.html?preset=subduction

The temperature/pressure tool now actually displays temperature and pressure values. Since there aren't currently any temperature/pressure values in the model this PR just uses faux temperature/pressure values that are proportional to the Y value in the cross-section. ~~With this implementation the semitransparent cursor is rendered using the system cursor, which means that the temperature/pressure display can lag behind a bit when the cursor movement is fast and/or the machine is slow. If we think this is undesirable or distracting with a little more work we can hide the system cursor and draw both pieces ourselves, which would mean that both would lag the actual mouse movement but at least one wouldn't lag relative to the other.~~ ⬅️ This has been done now.


js/components/simulation.tsx
```diff
@@ -89,6 +92,10 @@ export default class Simulation extends BaseComponent<IBaseProps, IState> {
         <div className="cross-section-container">
           <CrossSection />
         </div>
+        {
+          isMeasuringTempPressure &&
+          <TempPressureOverlay simulationStore={this.simulationStore} />
```
pjanik, line 97: If you need to pass `simulationStore`, I guess you might want to use:
```javascript
@inject("simulationStore")
@observer
```
in `TempPressureOverlay`. It's only one level between the parent and child component, so it doesn't change that much. But it's gonna be more consistent with the rest of the app.

pjanik (changes requested):
It works nicely. I've left one inline comment about using `@inject` and `@observer`. I think that's important, as AFAIR `@observer` guarantees that the component is re-rendered when necessary (when observable property is updated, like `measuredTemp`). Things seem to work without that right now (not sure why?), but I think it's safer to follow the pattern and be consistent with rest of the app. This might help/prevent some issues if we migrate MobX to a newer version one day.

> With this implementation the semitransparent cursor is rendered using the system cursor, which means that the temperature/pressure display can lag behind a bit when the cursor movement is fast and/or the machine is slow. If we think this is undesirable or distracting with a little more work we can hide the system cursor and draw both pieces ourselves, which would mean that both would lag the actual mouse movement but at least one wouldn't lag relative to the other.

I think the way you implemented it is a better approach. At least the pointer doesn't lag and it's easier to aim and pick rock / area you want. Handling a lagging cursor might be a much worse experience for users.

On a related note - I was quite surprised to notice that the display wasn't very smooth on my very fast machine (M1 Pro). I think the main culprit is the `throttle` function and 17ms throttle. My laptop display is 120Hz and when I set the throttle value to `1000 / 120` it felt much better. I didn't test 16ms which might be a bit better value too (since 17ms theoretically might skip some frames).

I think the best approach here could be to handle a mouse event in a requestAnimationFrame handler. We could just save the last mouse event in a React.ref quickly, and requestAnimationFrame callback would deal with the current ref value and update state. We'd avoid `throttle` at all and we'd sync with user display refresh rate automatically. But this is high level of fine tuning, so it doesn't need to happen now (or at all).

Also, I've noticed a funny effect when you actually try to click the cross-section and move your cursor:
https://user-images.githubusercontent.com/767857/165206670-b8e9155a-96a2-4d52-bb8f-679550fb121e.mov
Again, this is a minor detail and it doesn't need to be fixed now or at all.

kswenson (commented):
>I think the best approach here could be to handle a mouse event in a requestAnimationFrame handler. We could just save the last mouse event in a React.ref quickly, and requestAnimationFrame callback would deal with the current ref value and update state. We'd avoid throttle at all and we'd sync with user display refresh rate automatically. But this is high level of fine tuning, so it doesn't need to happen now (or at all).

Good idea -- done.

>Also, I've noticed a funny effect when you actually try to click the cross-section and move your cursor.... Again, this is a minor detail and it doesn't need to be fixed now or at all.

I took a quick look at this and didn't see an easy fix so I'm punting this one for later.

pjanik (approved):
Looks good, thanks. It works much more smooth on my laptop now.

> To use the decorators I'd have to switch to a class component, but you're right about adding observer. Will be in the next commit.

My preference would be to either stick to class components and decorators. Then, we're consistent across the app, and when we want to migrate to a new MobX / pattern, at least we precisely know what we need to update. This component will be updated halfway, and might get lost during refactoring.

Or... another, probably better, option would be to start a new pattern, but make it complete, and pass the store using a more convenient way than in props. I think passing down the store in props won't be great for most of the other components and somewhat defeats the purpose of using libraries like MobX. Here it's fine, as it's only one level down.

The new versions (mobx-react-lite if I'm correct) recommend using `observer` as you did, and the store can be retrieved from context / hook.

The hook and context is not much work. My concern was whether we can easily mix new and old MobX versions, but it seems Wildfire Model does it:
- component with decorators: https://github.com/concord-consortium/wildfire-model/blob/master/src/components/bottom-bar.tsx
- component with `useStores` hook: https://github.com/concord-consortium/wildfire-model/blob/master/src/components/view-3d/spark.tsx
- `useStores` hook itself: https://github.com/concord-consortium/wildfire-model/blob/master/src/use-stores.ts

Having said all that - I don't think this needs to be part of this PR. 😉 But I think it'd be nice to start this new pattern before we add too many new, functional components.

PR #210, opened 2022-04-28
Calculate "real" pressure & temperature

pjanik:
[#181934931]

Model rules are based on:
https://docs.google.com/presentation/d/1akzkA5gWRHQ-ZjPFrrlOjZvNfYoPorTh0r149yNp7so/edit#slide=id.p
Some slides are not matching the others, so I've just tried to do my best to average all that. I'll let Chris Lore just play with the implementation and tweak it more if necessary.


kswenson (approved):
👍 Looks good!

PR #211, opened 2022-04-28
Add divergent boundary magma animation

pjanik:
https://www.pivotaltracker.com/story/show/180800005

This is an attempt to add divergent boundary magma animation. I bet the project team might request more changes, but let's see.

https://tectonic-explorer.concord.org/branch/magma-animation/index.html

I bet you might not notice different frames of magma animation. I think the main reason is a pretty low resolution of cross-section (we could try to increase it, it's doable, but it'll have performance cost), small size of the magma image/animation, pretty subtle changes between frames, and the fact that its bouncing left and right so it's not easy to follow (I don't think there's a reasonably easy way to change it, as that's deep in the model physics / rules).

I've added a new URL option that lets you configure how often we update the cross-section view. A lower value will make the cross-section animation much smoother, but also it'll use more resources:
- https://tectonic-explorer.concord.org/branch/magma-animation/index.html?crossSectionUpdateInterval=3
- https://tectonic-explorer.concord.org/branch/magma-animation/index.html?crossSectionUpdateInterval=1


js/config.ts
```diff
@@ -177,7 +177,12 @@ const DEFAULT_CONFIG = {
   debug: false, // Adds features specific for debugging or model development, e.g. plates visibility toggle
   debugCrossSection: false,
   benchmark: false,
-  markCrossSectionFields: false
+  markCrossSectionFields: false,
+  // How often model web worker should send a message with fields data to the main UI thread.
+  fieldsUpdateInterval: 10,
+  // How often model web worker should send a message with cross-section data to the main UI thread.
+  crossSectionUpdateInterval: 5,
```
kswenson, line 184: In what unit?

kswenson (approved):
👍 Code looks fine -- simpler than I expected. I'll leave others to review the resulting effect. 😉

js/config.ts
pjanik: 10 means that fields are updated every 10th message sent from worker to UI (and that depends on machine performance, but the target is 30 fps / model steps per second). So, the lower, the more often. I can add comments about that.

PR #212, opened 2022-04-29
Downgrade screenfull to the last non-ESM version to avoid problems in older browsers.

pjanik:
https://www.pivotaltracker.com/story/show/182036906

The last screenful version mentions in their readme:

> This package is ESM. Please [familiarize](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) yourself with that that implies.
If you cannot use ESM or need to support older browsers without using transpilation, use version 5.2.0.

Chris Lore sent me a screenshot from old Chromebook and it confirms the problem is in the screenfull package source.

I should probably learn more about ESM and its implications and consider some other options. But this looks like a quick and safe fix for now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "tectonic-explorer",
-  "version": "2.1.0",
+  "version": "2.2.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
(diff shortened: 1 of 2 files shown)


kswenson (approved):
👍 LGTM

PR #215, opened 2022-05-02
Change polar cap boundary highlighting

kswenson:
PT: [[#180335792]](https://www.pivotaltracker.com/story/show/180335792)
Demo: https://tectonic-explorer.concord.org/branch/polar-boundary/index.html

### DIFF
--- a/js/stores/helpers/boundary-utils.ts
+++ b/js/stores/helpers/boundary-utils.ts
@@ -58,14 +58,58 @@ export function setHighlightStartingFromField(field: FieldStore, model: ModelSto
   }
 }
 
-export function highlightBoundarySegment(field: FieldStore, model: ModelStore): FieldStore[] {
+export function highlightPolarCapBoundary(field: FieldStore, model: ModelStore) {
+  const neighborPlateFields = new Map<number, FieldStore>();
+  const stack: FieldStore[] = [field];
+  while (stack.length > 0) {
+    const f = stack.pop() as FieldStore;
+    if (!f.highlighted) {
+      f.highlighted = true;
+    }
+
+    // identify the boundary fields of the polar cap plate
+    f.forEachNeighbor(n => {
+      if (!n.highlighted && n.boundary && (n.plate.id === field.plate.id)) {
+        stack.push(n);
+      }
+    });
+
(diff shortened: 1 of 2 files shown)


js/stores/helpers/boundary-utils.ts
```diff
@@ -58,14 +58,58 @@ export function setHighlightStartingFromField(field: FieldStore, model: ModelSto
...
-    return [field, fieldFromNeighboringPlate];
+    highlighted = [field, fieldFromNeighboringPlate];
+    const polarCapField = field.plate.isPolarCap
+                            ? field
+                            : fieldFromNeighboringPlate.plate.isPolarCap
+                                ? fieldFromNeighboringPlate
+                                : undefined;
+    if (polarCapField) {
+      // polar cap plates are highlighted all the way around
+      highlighted = [...highlighted, ...highlightPolarCapBoundary(polarCapField, model)];
```
pjanik, line 105: I think the `highlighted` array will have duplicate values. It should not cause any problems, as it it's only used to re-render plates. But I guess it could be easy to fix just to prevent some future issues if we start using this array more / for other things.

And an opinionated comment - while this one line expression is neat and small, it wasn't obvious to me what `highlightPolarCapBoundary` was returning (as it clearly has two separate goals - highlighting and returning neighbors). Maybe a comment or two lines could probably improve that:
```javascript
const polarCapPlateNeighbors = highlightPolarCapBoundary(polarCapField, model)
highlighted = [...highlighted, ...polarCapPlateNeighbors];
```

pjanik (approved):
It looks good to me 👍 

So, I bet your issue that we talked about was related to plate re-rendering, right? Sorry I didn't catch it.

I've left some inline comments, but nothing critical.

kswenson:
> So, I bet your issue that we talked about was related to plate re-rendering, right? Sorry I didn't catch it.

Yes, I spent a while debugging presumed problems in my highlighting code before figuring out that the highlighting code was working fine but all of the necessary plates weren't re-rendering.

js/stores/helpers/boundary-utils.ts
kswenson: Done and done

PR #219, opened 2022-05-03
Implement geode/tecrocks mode wizard [#181065898]

kswenson:
PT: [[#181065898]](https://www.pivotaltracker.com/story/show/181065898)
Demo: https://tectonic-explorer.concord.org/branch/mode-wizard/index.html


pjanik (approved):
It looks good. 👍 

It isn't fully related to this PR as we had this URL param implemented already, but I'm wondering if it's not a good moment to come up with a better name for it.

`geode=true` used to be used only by LARA / AP activities with the old model, so it was never visible for users. Now, some people might try to remember it or bookmark it. I'm wondering if some name like `rocks=true/false` wouldn't be more meaningful (and we could still support `geode` so old activities don't have to be updated). Not sure if it matters tho, and I guess it doesn't have to be part of this PR.

pjanik (approved):
Looks good, thanks.

kswenson:
@pjanik One more wrinkle -- project team wants the reload button (the internal one on the top bar) to return to the mode selection screen.

pjanik (approved):
Looks good 👍

PR #221, opened 2022-05-04
Implement block faulting around cont-cont divergent boundaries

pjanik:
[#177579801]

Demo: https://tectonic-explorer.concord.org/branch/block-faulting/index.html?modelId=07823204-632d-4b05-8712-7f055b372e3a&rocks=true

This is as close as I could get to some drafts you posted in the PT story. The diagonal lines are fake, they just go within one column / field, but I think the visual effect works okay. The code turned out to be pretty small / simpler than expected (it also shows a complete round trip of data, from model to cross-section rendering code).

You can also change how strong is the faulting using a new url param:
- stronger: [https://tectonic-explorer.concord.org/branch/block-faulting/index.html?modelId=07823204-632[…]8712-7f055b372e3a&rocks=true&blockFaultingStrength=0.3](https://tectonic-explorer.concord.org/branch/block-faulting/index.html?modelId=07823204-632d-4b05-8712-7f055b372e3a&rocks=true&blockFaultingStrength=0.3)
- weaker: [https://tectonic-explorer.concord.org/branch/block-faulting/index.html?modelId=07823204-632[…]8712-7f055b372e3a&rocks=true&blockFaultingStrength=0.1](https://tectonic-explorer.concord.org/branch/block-faulting/index.html?modelId=07823204-632d-4b05-8712-7f055b372e3a&rocks=true&blockFaultingStrength=0.1)
- disabled: [https://tectonic-explorer.concord.org/branch/block-faulting/index.html?modelId=07823204-632[…]5-8712-7f055b372e3a&rocks=true&blockFaultingStrength=0](https://tectonic-explorer.concord.org/branch/block-faulting/index.html?modelId=07823204-632d-4b05-8712-7f055b372e3a&rocks=true&blockFaultingStrength=0)

### DIFF
--- a/js/plates-model/model.ts
+++ b/js/plates-model/model.ts
@@ -334,6 +334,9 @@ export default class Model {
                 props.age = neighbor.age;
                 // When continent is being stretched, move field marker to the new field to emphasise this effect.
                 props.marked = neighbor.marked;
+                // `blockFaulting` value doesn't have physical meaning, but it's used to determine its direction in the
+                // rendering code. 1e6 value is big enough to cover all the visible fields in the cross-section.
+                props.blockFaulting = (neighbor.blockFaulting ?? 1e6) - 1;
                 neighbor.marked = false;
               } else {
                 props.type = "ocean";
--- a/js/config.ts
+++ b/js/config.ts
@@ -92,6 +92,8 @@ const DEFAULT_CONFIG = {
   continentDensity: 3,
   // How fast continent is stretching along divergent boundary. Bigger value means it would turn into ocean / sea faster.
   continentalStretchingRatio: 4,
+  // Intensity of the block faulting around divergent continent-continent boundary. 0 will turn it off completely.
+  blockFaultingStrength: 0.2,
   // Max length of the cross-section line
   maxCrossSectionLength: 4000, // km
   // Horizontal scaling of cross-section data.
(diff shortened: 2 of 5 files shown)


js/plates-model/model.ts
```diff
@@ -334,6 +334,9 @@ export default class Model {
                 props.age = neighbor.age;
                 // When continent is being stretched, move field marker to the new field to emphasise this effect.
                 props.marked = neighbor.marked;
+                // `blockFaulting` value doesn't have physical meaning, but it's used to determine its direction in the
+                // rendering code. 1e6 value is big enough to cover all the visible fields in the cross-section.
+                props.blockFaulting = (neighbor.blockFaulting ?? 1e6) - 1;
```
kswenson, line 339: `1e6 - 1` looks a little odd here, but just to confirm my understanding, the actual number doesn't matter since the only use for this value is the comparison down below, so the only thing that matters is which value is the relative value between neighbors.

kswenson (approved):
👍 Looks good -- very elegant!

js/plates-model/model.ts
pjanik: > the only thing that matters is the relative value between neighbors.

Yes, exactly. I've used 1e6, as it's pretty safe we won't have more than 1 million fields in a row :wink: Each next field will have a value decreased by 1. And even if we ever reach 0 for whatever reason, it'll actually automatically restart counting and it won't have big implications.

PR #222, opened 2022-05-04
Add thin layer of sediment to surface of continents [#181799207]

kswenson:
PT: [[#181799207]](https://www.pivotaltracker.com/story/show/181799207)
Demo: https://tectonic-explorer.concord.org/branch/continental-sediment/index.html?rocks

### DIFF
--- a/js/plates-model/crust.ts
+++ b/js/plates-model/crust.ts
@@ -207,18 +207,20 @@ export default class Crust {
         { rock: Rock.Gabbro, thickness: thickness * 0.7 }
       ];
     } else if (fieldType === "continent") {
-      let continentalSediment: Rock;
+      let continentalCrust: Rock;
       const range = BASE_CONTINENTAL_CRUST_THICKNESS - BASE_OCEANIC_CRUST_THICKNESS;
       if (thickness < BASE_OCEANIC_CRUST_THICKNESS + 0.4 * range) {
-        continentalSediment = Rock.Limestone;
+        continentalCrust = Rock.Limestone;
       } else if (thickness < BASE_OCEANIC_CRUST_THICKNESS + 0.8 * range) {
-        continentalSediment = Rock.Shale;
+        continentalCrust = Rock.Shale;
       } else {
-        continentalSediment = Rock.Sandstone;
+        continentalCrust = Rock.Sandstone;
       }
+      const granite = Math.max(0, thickness - SHALE_LIMESTONE_SANDSTONE_THICKNESS - MAX_REGULAR_SEDIMENT_THICKNESS);
       this.rockLayers = [
-        { rock: continentalSediment, thickness: SHALE_LIMESTONE_SANDSTONE_THICKNESS },
-        { rock: Rock.Granite, thickness: thickness - SHALE_LIMESTONE_SANDSTONE_THICKNESS }
+        { rock: Rock.ContinentalSediment, thickness: MAX_REGULAR_SEDIMENT_THICKNESS },
+        { rock: continentalCrust, thickness: SHALE_LIMESTONE_SANDSTONE_THICKNESS },
+        { rock: Rock.Granite, thickness: granite }
       ];
     } else if (fieldType === "island") {
       const oceanicBaseThickness = Math.min(thickness, BASE_OCEANIC_CRUST_THICKNESS);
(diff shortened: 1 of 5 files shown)


js/plates-model/rock-properties.ts
```diff
@@ -25,6 +26,12 @@ export interface IRockProperties {
 }
 
 export const ROCK_PROPERTIES: Record<Rock, IRockProperties> = {
+  [Rock.ContinentalSediment]: {
+    label: "Continental Sediments",
+    orderIndex: 0,
```
kswenson, line 31: Guessing at the order here, but presumably continental and oceanic sediments don't interact much. If they were to overlap, we'll put continental on top.

js/plates-model/rock-properties.ts
```diff
@@ -25,6 +26,12 @@ export interface IRockProperties {
...
+    isTransferrableDuringCollision: true,
+    canSubduct: true
```
kswenson, line 33: PT story says "The continental sediment will react to changing P/T conditions as any other rock would, treat it like oceanic sediments."

js/plates-view/plate-mesh.ts
```diff
@@ -73,6 +74,7 @@ const ROCK_PATTERN_SCALE: Record<Rock, number> = {
   [Rock.Rhyolite]: 30,
   [Rock.Andesite]: 120,
   [Rock.Diorite]: 1,  // never visible on surface
+  [Rock.ContinentalSediment]: 30,
```
kswenson, line 77: Same value as for oceanic sediment.

js/plates-model/crust.ts
```diff
@@ -207,18 +207,20 @@ export default class Crust {
...
+        continentalCrust = Rock.Shale;
       } else {
-        continentalSediment = Rock.Sandstone;
+        continentalCrust = Rock.Sandstone;
       }
+      const granite = Math.max(0, thickness - SHALE_LIMESTONE_SANDSTONE_THICKNESS - MAX_REGULAR_SEDIMENT_THICKNESS);
       this.rockLayers = [
-        { rock: continentalSediment, thickness: SHALE_LIMESTONE_SANDSTONE_THICKNESS },
-        { rock: Rock.Granite, thickness: thickness - SHALE_LIMESTONE_SANDSTONE_THICKNESS }
+        { rock: Rock.ContinentalSediment, thickness: MAX_REGULAR_SEDIMENT_THICKNESS },
```
kswenson, line 221: PT story says "about as much as the oceanic sediment in the oceans" so we reuse the same sediment thickness constant.

pjanik: I'd still set `canSubduct` to `false`. This property decides whether crust/field can subduct or not. And we assume that only oceanic fields can subduct. That's why only oceanic rocks have this property set to true.

In other words, you could think that canSubduct is a property that will decide whether a crust is oceanic or continental. The name is a bit more generic tho, as we also have islands, etc. Continental sediment is part of a continental crust.

pjanik (changes requested):
It looks good! 👍 

I've left one small comment about `canSubduct` value. It won't change anything in practice, as the continental crust will pretty much always have some other rocks that have this property set to false. But I'd still set it to `false` to be consistent and to follow the initial idea there. If this wasn't clear before, and now it is more, maybe you could add a comment that would make that clear for other devs too. Thanks!

js/plates-model/rock-properties.ts
pjanik: I'd still set canSubduct to false. This property decides whether crust/field can subduct or not. And we assume that only oceanic fields can subduct. That's why only oceanic rocks have this property set to true.

In other words, you could think that canSubduct is a property that will decide whether a crust is oceanic or continental. The name is a bit more generic tho, as we also have islands, etc. Continental sediment is part of a continental crust.

pjanik (changes requested):
It looks good! 👍 

I've left one small comment about `canSubduct` value. It won't change anything in practice, as the continental crust will pretty much always have some other rocks that have this property set to false. But I'd still set it to `false` to be consistent and to follow the initial idea there. If this wasn't clear before, and now it is more, maybe you could add a comment that would make that clear for other devs too. Thanks!

pjanik (approved):
LGTM 👍

PR #223, opened 2022-05-05
Add thrust faulting around convergent boundaries

pjanik:
https://www.pivotaltracker.com/story/show/181799189

Demo: https://tectonic-explorer.concord.org/branch/thrust-faulting/index.html?modelId=550f18df-ad6e-4714-bc24-f021c64c5f2d&rocks=true

This PR adds thrust faulting around convergent continent-continent boundaries. PT story has a bit unclear description (I guess copied over from divergent boundary version) and the draft image there is very optimistic. But this is what I could do for now. Let's see what's the project team feedback.

### DIFF
--- a/js/plates-model/field.ts
+++ b/js/plates-model/field.ts
@@ -432,7 +432,7 @@ export default class Field extends FieldBase<Field> {
     this.forEachNeighbor((n) => {
       if (n.oceanicCrust && !n.subduction && n.bendingProgress < possibleNeighBending &&
         // This line below checks if vector that connects neighboring field with this one is pointing the opposite
-        // direction than the velocity the (subducting) plate. This ensures that plate bending progresses
+        // direction than the velocity of the (subducting) plate. This ensures that plate bending progresses
         // against the plate movement direction. It makes sense and lets us avoid some unwanted effects. See:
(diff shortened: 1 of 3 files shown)


js/plates-model/field.ts
```diff
@@ -394,16 +398,29 @@ export default class Field extends FieldBase<Field> {
       this.crust.fold(timestep, neighboringCrust, this.colliding.subduction.relativeVelocity);
       this.crust.setMetamorphic(1);
       this.colliding.crust.setMetamorphic(1);
+      if (this.blockFaulting === 0) {
+        // Add block faulting to the top plate that is part of the continental collision. The actual value has no physical
+        // meaning, but larger value will propagate more (check #propagateBlockFaulting()) making the block faulting area
+        // wider. And the difference between between the two neighboring fields defines faulting direction in the rendering code.
```
kswenson, line 404:
```suggestion
        // wider. And the difference between the two neighboring fields defines faulting direction in the rendering code.
```

kswenson (approved):
👍 Code looks good and result looks good! I'm glad those turned out to be more straightforward than you'd/we'd feared.

PR #225, opened 2022-05-05
Tweak block faulting

pjanik:
[#177579801]
[#181799189]

The project team requested a few changes:
- Shallower block faulting lines. They mentioned 40km, but not sure what that means. I've made that configurable.
- Weaker faulting. I've updated the default config balue.
- No faulting in the subducting plate during orogeny / continental collision. It's removed now.
- Sequential propagation of the faulting in the top blat during orogeny / continental collision rather than immediate change. I've done it by simplifying the logic. I don't propagate `blockFaulting` value anymore. Instead, I've just set it using subducting progress of the plate underneath. This makes the faults appear in succession.


kswenson (approved):
👍 Remarkable how those changes made the code simpler!

PR #226, opened 2022-05-06
Extend range of the crust age visualization

pjanik:
https://www.pivotaltracker.com/story/show/180997501

This is only the beginning of the work. I've extended range of `normalizedAge` value from [0, 1] to [0, 5]. That let me easily show a wider range of crust age and it's more or less matching Michael's mocks. 
See: https://www.pivotaltracker.com/story/show/180997501/comments/230961040

`normalizedAge` is used in multiple places in the model, so I think it's important that this getter still behaves the same in [0,1] range. Multiple places in the model check if it's < 1, which seems fine here and saves us and makes implementation simpler.

A new property could also work, but it might be more confusing to have two separate concepts of age, more data to send between threads and to serialize.

I guess we could update the name of the getter. Suggestions are welcome. I haven't had a better idea, so I've left it. Maybe it's fine with the comments I've added. It's normalized up to some point. 😉 

So, the remaining work is to update colormap and deal with distinct pre-existing crust.

Also, the current age is based on traveled distance only. So, if the plates get merged together, we'll see the "fresh" crust coloring forever. I wonder if it should become dark blue / gray over time. If so, we should replace distance with passed time. In this case, it might require a new property.

I think the old age uses distance so the shape of the oceanic ridge doesn't depend on the speed of the plate. If it's based on time only, if the plate is moving slowly, the ridge will have a very gentle slope. If it's moving very fast, it'll have a very steep slope.

Old:
Screenshot 2022-05-05 at 20 18 42

New:
Screenshot 2022-05-05 at 20 18 10

### DIFF
--- a/js/plates-model/field.ts
+++ b/js/plates-model/field.ts
@@ -227,7 +227,7 @@ export default class Field extends FieldBase<Field> {
   // to crust age visualization required to extend it to larger values, so we show broader range of the fresh crust.
   // TODO: rename this property? It should not break serialization as long as `.age` is not changed.
   get normalizedAge() {
-    return this.age / MATURE_CRUST_AGE;
+    return Math.min(MAX_NORMALIZED_AGE, this.age / MATURE_CRUST_AGE);
   }
 
   get rockType() {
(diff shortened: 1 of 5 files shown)


kswenson (approved):
👍 Code changes look fine and make sense. I'll have the updated key available later today.

PR #230, opened 2022-05-07
Update crust age visualization

pjanik:
[#180997501]

This PR adds a new crust age visualization. It handles preexisting crust too.

Demo: https://tectonic-explorer.concord.org/branch/new-crust-age-vis/index.html?rocks=true&preset=divergentBoundary

### DIFF
--- a/js/plates-view/plate-mesh.ts
+++ b/js/plates-view/plate-mesh.ts
@@ -7,8 +7,8 @@ import TemporalEvents from "./temporal-events";
 import { earthquakeTexture, depthToColor, magnitudeToSize } from "./earthquake-helpers";
 import { volcanicEruptionTexture } from "./volcanic-eruption-helpers";
 import PlateLabel from "./plate-label";
-import { hueAndElevationToRgb, MAX_ELEVATION, MIN_ELEVATION, normalizeElevation, topoColor } from "../colors/topographic-colors";
-import { RGBAFloat, RGBAFloatToHexNumber } from "../colors/utils";
+import { crustAgeColor, hueAndElevationToRgb, MAX_ELEVATION, MIN_ELEVATION, normalizeElevation, preexistingCrustAgeColor, topoColor } from "../colors/topographic-colors";
+import { cssColorToRGBAFloat, RGBAFloat, RGBAFloatToHexNumber } from "../colors/utils";
 import config, { Colormap } from "../config";
 import getGrid from "../plates-model/grid";
 import { autorun, observe } from "mobx";
@@ -24,6 +24,7 @@ const MIN_SPEED_TO_RENDER_POLE = 0.002;
 const VELOCITY_ARROWS_DIVIDER = 3;
 const BOUNDARY_COLOR = { r: 0.8, g: 0.2, b: 0.5, a: 1 };
 const HIGHLIGHT_COLOR = { r: 0.38, g: 0.86, b: 0.04, a: 1 };
+const PREEXISTING_CRUST_COLOR = cssColorToRGBAFloat(preexistingCrustAgeColor);
 // Special color value that indicates that colormap texture should be used.
 const USE_COLORMAP_COLOR = { r: 0, g: 0, b: 0, a: 0 };
 // Larger value will make mountains in the 3D view more pronounced.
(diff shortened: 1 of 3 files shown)


js/plates-view/plate-mesh.ts
```diff
@@ -352,9 +351,10 @@ export default class PlateMesh {
     } else if (colormap === "age") {
       this.material.uniforms.usePatterns.value = false;
       this.material.uniforms.colormap.value = this.colormapTextures.age;
-      // Fields at the divergent boundary should have 0 age.
-      this.defaultColormapValue = 0;
-      this.defaultColorAttr = { r: 0, g: 0, b: 0, a: 0 };
+      // Fields at the divergent boundary should have max age to avoid strange color transition around preexisting
+      // crust borders.
+      this.defaultColormapValue = 1;
+      this.defaultColorAttr = PREEXISTING_CRUST_COLOR;
```
pjanik, line 357: These two lines are actually very important. You can try to set `defaultColormapValue` to `0` and `defaultColorAttr` to any other color to see the difference.

This might become important if Amy or Chris complain about this small artifact:
Screenshot 2022-05-06 at 20 05 03

However, I don't see a better solution for now. If we set defaultColormapValue to 0, then each plate will have yellow border which doesn't seem to make much sense.

js/plates-view/plate-mesh.ts
```diff
@@ -404,6 +404,16 @@ export default class PlateMesh {
...
+    if (this.store.colormap === "age" && field.normalizedAge === MAX_NORMALIZED_AGE) {
+      // Note that this is special case for preexisting crust. Field color attribute will be used instead of
+      // regular crust age colormap (it's only applied to "fresh" crust)
+      return PREEXISTING_CRUST_COLOR;
+    }
+    // If crust age colormap should not be used, this would be an option. It avoid some strange color transitions,
+    // but provides more blurry and les precise rendering.
+    // if (this.store.colormap === "age") {
+    // return crustAgeColor(field.normalizedAge);
+    // }
```
pjanik, line 416: If you uncomment these three lines, we'll stick to vertex color attribute style of rendering. It can be useful to try to see and understand the difference (and/or show to the project team if they're not happy about the current result).

js/plates-view/plate-mesh.ts
```diff
@@ -434,14 +444,19 @@ export default class PlateMesh {
...
 
-    const colormap = this.store.colormap;
-    if (colormap === "topo" || colormap === "plate") {
-      this.geoAttributes.colormapValue.setX(id, normalizeElevation(field.elevation));
-    } else if (colormap === "age") {
-      this.geoAttributes.colormapValue.setX(id, field.normalizedAge / MAX_NORMALIZED_AGE);
-    } else if (colormap === "rock") {
-      // colormapValue will be used to pick the texture in plate-mesh-fragment.glsl.
-      this.geoAttributes.patternIdx.setX(id, ROCK_PATTERN_IDX[field.rockType] || 0);
+    if (color === USE_COLORMAP_COLOR) {
```
pjanik, line 447: I've added this line to make the flow a bit clearer / more understandable (and we will skip some operations that were not necessary).

js/plates-view/plate-mesh.ts
kswenson: I don't suppose there's some simple smoothing operation we could perform when rendering?

js/plates-view/plate-mesh.ts
```diff
@@ -404,6 +404,16 @@ export default class PlateMesh {
     if (this.store.renderBoundaries && field.boundary) {
       return BOUNDARY_COLOR;
     }
+    if (this.store.colormap === "age" && field.normalizedAge === MAX_NORMALIZED_AGE) {
+      // Note that this is special case for preexisting crust. Field color attribute will be used instead of
+      // regular crust age colormap (it's only applied to "fresh" crust)
+      return PREEXISTING_CRUST_COLOR;
+    }
+    // If crust age colormap should not be used, this would be an option. It avoid some strange color transitions,
+    // but provides more blurry and les precise rendering.
```
kswenson, line 413:
```suggestion
    // but provides more blurry and less precise rendering.
```

js/plates-view/plate-mesh.ts
kswenson: Or maybe combine the two in some interesting way?

kswenson (approved):
👍 Code looks good -- I'm glad you did it before you left. 😉  We'll have to wait to see what the project team thinks.

js/plates-view/plate-mesh.ts
pjanik: The problem here is that what you see between the plates is a special type of a field - `adjacentField`. It's a field that is not yet part of the model, but it's a "field candidate". It's added to the model when it can be (there's enough of gap between plates). I'm rendering these fields so there are no visible gaps between plates, plus they have a few more functions. And the mentioned lines affect their rendering properties as far as I remember.

PR #233, opened 2022-05-11
Implement zoom control buttons for cross-section [#181909786]

kswenson:
PT: [[#181909786]](https://www.pivotaltracker.com/story/show/181909786)
Demo: https://tectonic-explorer.concord.org/branch/zoom-buttons/index.html?preset=benchmark&rocks


js/components/cross-section-3d.tsx
```diff
@@ -57,16 +67,10 @@ export default class CrossSection3D extends BaseComponent<IBaseProps, IState> {
...
 
   render() {
-    const { showCrossSectionCameraReset, resetCrossSectionCamera, interaction, key } = this.simulationStore;
+    const { interaction, key } = this.simulationStore;
     return (
       <div className={`cross-section-3d-view ${interaction} ${key ? "narrow" : ""}`} data-test="3D-view">
         <div ref={(c) => this.view3dContainer = c} />
-        {
-          showCrossSectionCameraReset &&
-          <SmallButton className="cross-section-camera-reset" onClick={resetCrossSectionCamera} icon="settings_backup_restore" data-test="camera-reset">
```
pjanik, line 66: Probably related CSS can be removed.

js/components/cross-section.tsx
```diff
@@ -1,32 +1,67 @@
...
+  handleZoomOut = () => {
+    if (this.camera) {
+      this.camera.zoom = Math.max(MIN_CAMERA_ZOOM, this.camera.zoom - CAMERA_ZOOM_STEP);
+      this.camera.updateProjectionMatrix();
+      this.setState({ cameraZoom: this.camera.zoom });
+    }
+  };
+
+  handleResetCamera = () => {
+    this.camera.zoom = 1;
```
pjanik, line 46: It's a minor issue, but it's a bit harsh compared to the smooth camera angle animation. 😉

js/components/cross-section.tsx
```diff
@@ -1,32 +1,67 @@
...
+    this.camera.updateProjectionMatrix();
+    this.simulationStore.resetCrossSectionCamera();
+    this.setState({ cameraZoom: this.camera.zoom });
+  };
+
   render() {
-    const { crossSectionVisible, closeCrossSection } = this.simulationStore;
+    const { crossSectionVisible, showCrossSectionCameraReset, closeCrossSection } = this.simulationStore;
+    const showResetCamera = (this.camera && this.camera.zoom !== 1) || showCrossSectionCameraReset;
```
pjanik, line 54: I think that `this.camera.zoom` shouldn't be used in `render()` function, as it isn't React state or MobX observable property. It'll work when `render()` is called for some reason, but it will never itself trigger re-render (and we could expect that looking at this logic).

If you mean using `this.state.zoom`, it'd be more correct, but I think you never update `this.state.zoom` on zoom change initiated by mouse wheel / scrolling.

I think it actually works, as `simulationStore.showCrossSectionCameraReset` becomes true when you zoom in or out. It was supposed to detect only camera angle change, but apparently it also detects zoom by accident (I've checked that in `master` branch).

pjanik (changes requested):
Generally, it works nicely. However, there are a few small issues that would be nice to fix if possible. I've mentioned them in inline comments.

Also, this is not related to the buttons themselves, but since you were working on related code, including saving `zoom` value in a component state, I think it could be a good moment to make it consistent with the cross-section angle value.

Note that `crossSectionCameraAngle` is stored in the simulation store, serialized, and deserialized. I guess it'd be nice if we did the same for the zoom. Then, we wouldn't have to store `zoom` in the CrossSection component state. I guess we could directly follow / copy patterns set by `crossSectionCameraAngle` to implement zoom observing and updating.

But if you think it's not related to the zoom buttons and it's not the best moment for this work, I guess it could be done in a separate PR.

js/components/cross-section.tsx
kswenson: In this implementation, `this.camera.zoom` is the ultimate source of truth, but `this.state.zoom` is updated as a means of triggering re-render. It was admittedly a rather quick-and-dirty solution but putting it in the `simulationStore` as you suggested above is probably the better solution.

js/components/cross-section.tsx
kswenson: True, but I wasn't sure it was worth the effort to implement the animation. It didn't seem to look _too_ bad when I was testing it.

pjanik (approved):
It looks and works great, thanks!

PR #232, opened 2022-05-10
Update magma animations

kswenson:
Divergent Boundary
PT: [[#180800005]](https://www.pivotaltracker.com/story/show/180800005)
Demo: https://tectonic-explorer.concord.org/branch/magma-animation/index.html?preset=divergentBoundary&rocks

Subduction
PT: [[#179785717]](https://www.pivotaltracker.com/story/show/179785717) [[#182106405]](https://www.pivotaltracker.com/story/show/182106405)
Demo: https://tectonic-explorer.concord.org/branch/magma-animation/index.html?preset=continentalCollision1&rocks

PR #234, opened 2022-05-16
Render divergent boundary magma after all fields

kswenson:
In #232, @pjanik wrote:

>The new magma image is wider than a field column, so its right edge might be cut by the rendering of the next field. If we render the image after rendering all the fields, it should fix it.


pjanik (approved):
Looks good 👍

PR #235, opened 2022-05-17
Update plate color key [#180992870]

kswenson:
PT: [[#180992870]](https://www.pivotaltracker.com/story/show/180992870)
Demo: https://tectonic-explorer.concord.org/branch/plate-color/index.html?rocks&preset=benchmark

### DIFF
--- a/js/components/keys/plate-color-key.tsx
+++ b/js/components/keys/plate-color-key.tsx
@@ -1,3 +1,4 @@
+import { observer } from "mobx-react";
 import React from "react";
 import ModelStore from "../../stores/model-store";
 import Plate1Color from "../../../images/plate-1-color.svg";
@@ -21,7 +22,7 @@ interface IProps {
   model: ModelStore;
 }
 
-export const PlateColorKey = ({ model }: IProps) => {
+export const PlateColorKey = observer(({ model }: IProps) => {
   const plateColors = plateCountToPlateColors[model.plates.length - 1];
   return (
     <div className={`${css.mapType} ${css.plateColorKey}`} data-test="map-type-key">
@@ -34,7 +35,7 @@ export const PlateColorKey = ({ model }: IProps) => {
       </div>
     </div>
   );
-};
+});
 
 const ElevationLabels = () => {
   const labels = ["8,000m\u00a0–", "4,000m\u00a0–", "0m\u00a0–", "–4,000m\u00a0–", "–8,000m\u00a0–"];
--- a/js/components/keys/plate-color-key.tsx
+++ b/js/components/keys/plate-color-key.tsx
@@ -1,0 +1,64 @@
+import React from "react";
+import ModelStore from "../../stores/model-store";
+import Plate1Color from "../../../images/plate-1-color.svg";
+import Plate2Color from "../../../images/plate-2-color.svg";
+import Plate2ColorAlt from "../../../images/plate-2-color-alt.svg";
+import Plate3Color from "../../../images/plate-3-color.svg";
+import Plate3ColorAlt from "../../../images/plate-3-color-alt.svg";
+import Plate4Color from "../../../images/plate-4-color.svg";
+import Plate5Color from "../../../images/plate-5-color.svg";
+
+import css from "../../../css-modules/keys/map-type.less";
+
+const plateCountToPlateColors: any[][] = [
+  [Plate1Color],
+  [Plate1Color, Plate2Color],
+  [Plate1Color, Plate2ColorAlt, Plate3ColorAlt],
+  [Plate1Color, Plate2Color, Plate3Color, Plate4Color],
+  [Plate1Color, Plate2Color, Plate3Color, Plate4Color, Plate5Color],
+];
+interface IProps {
+  model: ModelStore;
+}
+
+export const PlateColorKey = ({ model }: IProps) => {
+  const plateColors = plateCountToPlateColors[model.plates.length - 1];
+  return (
+    <div className={`${css.mapType} ${css.plateColorKey}`} data-test="map-type-key">
+      <div className={css.plateColorTitle}>Key: Plate Color (with elevation)</div>
+      <div className={css.plateColorGradients}>
+        <ElevationLabels/>
+        { plateColors.map((PlateColorGradient, i) => (
+          <PlateColorGradient className={css.plateColorGradient} key={`key-plate-${i}`} />)) }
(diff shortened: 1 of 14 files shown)


js/components/keys/plate-color-key.tsx
```diff
@@ -0,0 +1,64 @@
...
+  [Plate1Color, Plate2Color],
+  [Plate1Color, Plate2ColorAlt, Plate3ColorAlt],
+  [Plate1Color, Plate2Color, Plate3Color, Plate4Color],
+  [Plate1Color, Plate2Color, Plate3Color, Plate4Color, Plate5Color],
+];
+interface IProps {
+  model: ModelStore;
+}
+
+export const PlateColorKey = ({ model }: IProps) => {
```
pjanik, line 24: This component will not update itself / re-render when number of plates is changed. You can see it here:
https://tectonic-explorer.concord.org/branch/plate-color/index.html?rocks&preset=plates2
These two plates gets merged relatively soon (no forces, no motion). You'll see either 1 or 2 plates in the key, depending on when you open it. Also, resetting a model won't reset the key.

I think it's again the problem with functional components vs MobX. Probably adding `observer` will fix that, but I'm worried about passing various observables via props directly. This slightly defeats MobX's advantage of isolated re-render of the children without re-rendering of the parent. 

This is definitely not a task for this PR, but I think it'd be great to setup a complete pattern for MobX and functional components ASAP. We're supposed to have a ~week for maintenance, so I guess it'd be the perfect time for that.

pjanik (changes requested):
Looks great, but there's an issue with reacting to model updates. More in the inline comment.

Also, before it's caught by QA - Michael's specs mention dimming plates that no longer exist instead of removing the bar completely. This doesn't seem to be implemented. But I think there are issues with this concept anyway. An assumption that we reuse only 5 colors in this constant order is probably not so well thought. If you try to imagine cases when we divide and merge plates multiple times, the key quickly wouldn't match the model.

This is not a problem of this implementation, but probably something to discuss with Amy and Michael.

The last thing is the alternating colors - I guess Michael followed a pattern of existing presets thinking it's gonna help us. But I guess it was probably random.

kswenson:
> Also, before it's caught by QA - Michael's specs mention dimming plates that no longer exist instead of removing the bar completely. This doesn't seem to be implemented. But I think there are issues with this concept anyway. An assumption that we reuse only 5 colors in this constant order is probably not so well thought. If you try to imagine cases when we divide and merge plates multiple times, the key quickly wouldn't match the model.

I was figuring that would be part of the model work. Presumably we'd have to track the initial number of plates and the identity of each plate so that as plates merged/split we could determine which plate colors were no longer in use and thus which ones to dim. But I agree with your question in slack that this might not behave as intuitively as expected.

>This component will not update itself / re-render when number of plates is changed.

🤦‍♂️ Clearly, I was more focused on getting the appearance right. I've added the necessary `observer()`.

>I think it's again the problem with functional components vs MobX. Probably adding observer will fix that, but I'm worried about passing various observables via props directly. This slightly defeats MobX's advantage of isolated re-render of the children without re-rendering of the parent.
>
>This is definitely not a task for this PR, but I think it'd be great to setup a complete pattern for MobX and functional components ASAP. We're supposed to have a ~week for maintenance, so I guess it'd be the perfect time for that.

I'm happy to consider adopting a larger strategy for MobX and functional components, but in the cases in which it has come up so far, the functional component needed a single object that the parent component already possessed. In a React application, props are the right and proper way to pass information from a parent component to a child component. I believe the canonical solution for MobX/functional components is React context, which is a fine solution for avoiding prop-drilling but seems like overkill if we just need to get a single object from a parent to a child component.

pjanik:
> In a React application, props are the right and proper way to pass information from a parent component to a child component.

Right, but this has obvious limitations both in developer annoyance (typing and passing prop through multi-levels of components, often unrelated to the prop), and performance - to re-render the leaf of the component tree, we essentially need to re-render the whole tree. 

I believe MobX, Redux, and similar libraries solve both problems, and I think they're equally important in larger apps. Injection of the store lets us re-render the leaf only. Not sure if TecRocks is large enough and if it's premature optimization or not, but I think these small differences can add up. Of course I'm not saying that this particular key component will make any difference, but more about the whole app with passing props vs injection/context. I've implemented the context-based approach in one of the Geo projects and it was actually very simple. It's a few lines of code. And I think there's no alternative in the recent MobX versions (inject annotation is dead), but I haven't checked the docs recently.

pjanik (approved):

kswenson:
> > In a React application, props are the right and proper way to pass information from a parent component to a child component.
> 
> Right, but this has obvious limitations both in developer annoyance (typing and passing prop through multi-levels of components, often unrelated to the prop), and performance - to re-render the leaf of the component tree, we essentially need to re-render the whole tree.
> 
> I believe MobX, Redux, and similar libraries solve both problems, and I think they're equally important in larger apps. Injection of the store lets us re-render the leaf only. Not sure if TecRocks is large enough and if it's premature optimization or not, but I think these small differences can add up. Of course I'm not saying that this particular key component will make any difference (plus model is used by the parent anyway, as you said), but more about the whole app with passing props vs injection/context. I've implemented the context-based approach in one of the Geo projects and it was actually very simple. It's a few lines of code. And I think there's no alternative in the recent MobX versions (inject annotation is dead), but I haven't checked the docs recently.

I think we're saying essentially the same thing but viewing it from different perspectives.

>Right, but this has obvious limitations both in developer annoyance (typing and passing prop through multi-levels of components, often unrelated to the prop), and performance - to re-render the leaf of the component tree, we essentially need to re-render the whole tree.

These are real issues that have nothing to do with MobX -- if you need to get information from component A to component D via components B and C then you're prop-drilling and React context is a good solution to that problem. If you need to get a piece of information from component A to component B then React context is an unnecessary complication.

>And I think there's no alternative in the recent MobX versions (inject annotation is dead), but I haven't checked the docs recently.

In MobX React they included the `Provider` and `@inject` mechanisms along with `observer()` because they viewed it as their responsibility to have a solution for both the information provider and the consumer, in part because the React solutions to those problems weren't settled. With MobX React Lite they've dropped the provider-side support, I believe, because they've concluded that existing React mechanisms (e.g. props and context) are now sufficient to handle the provider side and so need not be implemented in MobX. In other words, decisions about whether to use context can be made for traditional React reasons because MobX needn't be part of that conversation.

PR #236, opened 2022-05-17
Implement new elevation (topographic) key design [#180992693]

kswenson:
PT: [[#180992693]](https://www.pivotaltracker.com/story/show/180992693)
Demo: https://tectonic-explorer.concord.org/branch/elevation-key/index.html?rocks&preset=benchmark

Changes look more extensive than they really are due to moving `ElevationLabels` and `ElevationDescriptions` components out of the `PlateColorKey` component and into their own components since they're now shared with the new `ElevationKey` component. Since the two keys are different heights the layout code computes the vertical position of the labels in JavaScript rather than in CSS.

### DIFF
--- a/css-modules/keys/elevation-descriptions.less
+++ b/css-modules/keys/elevation-descriptions.less
@@ -1,0 +1,34 @@
+.elevationDescriptions {
+  position: relative;
+  width: 96px;
+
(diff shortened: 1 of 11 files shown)


pjanik (approved):
Looks good, thanks 👍

PR #237, opened 2022-05-19
Update divergent boundary fresh crust overlay

pjanik:
[[#182215316]](https://www.pivotaltracker.com/story/show/182215316)

Another attempt to div boundary rendering. I demoed that to Chris and Trudi today during the meeting about plate colors, and they seemed to be okay with this version.

### DIFF
--- a/js/colors/rock-colors.ts
+++ b/js/colors/rock-colors.ts
@@ -81,7 +81,7 @@ preprocessRockPatterns();
 
 export const getRockPatternImgSrc = (rock: PatternName): string => ROCK_PATTERN[rock].patternImgSrc;
 
-export const getRockCanvasPattern = (ctx: CanvasRenderingContext2D, rock: PatternName) => {
+export const getRockCanvasPattern = (ctx: CanvasRenderingContext2D, rock: PatternName, age: number | undefined = undefined) => {
   const pattern = ROCK_PATTERN[rock];
   let canvasPattern = null;
   if (pattern.patternImgSrc !== "" && pattern.imgElement.complete) {
@@ -90,6 +90,10 @@ export const getRockCanvasPattern = (ctx: CanvasRenderingContext2D, rock: Patter
   // #ccc might be visible for a short moment while the pattern is still loading. However, given the pattern
   // image size, this is not likely to happen.
   return canvasPattern || "#ccc";
+};
+
+export const getRockCanvasPatternGivenNormalizedAge = (ctx: CanvasRenderingContext2D, rock: PatternName, normalizedAge: number) => {
+  return normalizedAge < 0.8 ? "#333" : getRockCanvasPattern(ctx, rock);
 };
 
 export const IGNEOUS_PURPLE = "#BA00BA";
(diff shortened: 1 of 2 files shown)


js/plates-view/render-cross-section.ts
```diff
@@ -424,7 +418,8 @@ class CrossSectionRenderer {
       const p2tmp = p2.clone().lerp(p3, currentThickness2);
       const p3tmp = p2.clone().lerp(p3, currentThickness2 + rl.relativeThickness2);
       const p4tmp = p1.clone().lerp(p4, currentThickness1 + rl.relativeThickness1);
-      if (this.fillPath(getRockCanvasPattern(this.ctx, rl.rock), p1tmp, p2tmp, p3tmp, p4tmp)) {
+      const color = (field.normalizedAge || 0) < 0.8 ? "#333" : getRockCanvasPattern(this.ctx, rl.rock);
```
kswenson, line 421: Given that this same expression occurs here and in line 399 above, you could DRY up the code a bit with a function like `getRockCanvasPatternGivenNormalizedAge()` but it's not a big deal either way.

kswenson (approved):
👍 Looks good!

js/plates-view/render-cross-section.ts
pjanik: Done in the last commit, thanks.

js/colors/rock-colors.ts
```diff
@@ -92,6 +92,10 @@ export const getRockCanvasPattern = (ctx: CanvasRenderingContext2D, rock: Patter
   return canvasPattern || "#ccc";
 };
 
+export const getRockCanvasPatternGivenNormalizedAge = (ctx: CanvasRenderingContext2D, rock: PatternName, normalizedAge: number) => {
```
kswenson, line 95: Could also use:
```
export const getRockCanvasPatternGivenNormalizedAge = (ctx: CanvasRenderingContext2D, rock: PatternName, normalizedAge = 0) => {
```
and then clients would not have to provide the `|| 0`.

js/colors/rock-colors.ts
```diff
@@ -81,7 +81,7 @@ preprocessRockPatterns();
 
 export const getRockPatternImgSrc = (rock: PatternName): string => ROCK_PATTERN[rock].patternImgSrc;
 
-export const getRockCanvasPattern = (ctx: CanvasRenderingContext2D, rock: PatternName) => {
+export const getRockCanvasPattern = (ctx: CanvasRenderingContext2D, rock: PatternName, age: number | undefined = undefined) => {
```
kswenson, line 84: Could also use:
```
export const getRockCanvasPattern = (ctx: CanvasRenderingContext2D, rock: PatternName, age?: number) => {
```

pjanik: This change actually should not have been committed at all, as it's not used. I've just removed it here: https://github.com/concord-consortium/tectonic-explorer/commit/099d94e79c90bd61faca283689753f4b6089d72d

js/colors/rock-colors.ts
pjanik: This was actually intentional, so if you try to get a rock pattern and don't think about age / normalized age, you don't try to use this function (and use `getRockCanvasPattern` instead). :wink: `|| 0` is never actually triggered in real life I believe, it's just a result of a loose interface.

PR #239, opened 2022-05-19
Fix cursor over zoom buttons [#181909786]

kswenson:
PT: [[#181909786]](https://www.pivotaltracker.com/story/show/181909786)


pjanik (approved):
Looks good 👍

PR #241, opened 2022-05-20
Updates related to new plate colors

pjanik:
The ideas have been changing a lot, but I've updated this story description to reflect the latest requirements:
https://www.pivotaltracker.com/story/show/182121571

There's some UI work and a few changes to model logic related to plate IDs, colors, and merge and division.

I've also merged into this branch changes from this PR: https://github.com/concord-consortium/tectonic-explorer/pull/238, as they're touching the same code.


js/plates-model/model.ts
```diff
@@ -77,10 +80,40 @@ export default class Model {
...
+      this.plates.forEach(p => {
+        availableHues.delete(p.hue);
+      });
+      return getRandomItem(availableHues);
+    }
+    // Very unlikely, but if all the available hues are used, return random color.
+    if (hueToAvoid !== undefined) {
+      availableHues.delete(hueToAvoid);
+    }
+    return getRandomItem(availableHues);
```
kswenson, line 114: Is random better than something like `availableHues[id % availableHues.length]`? Seems like there might be some value to a consistently repeating pattern.

kswenson (approved):
👍 Looks good!

js/plates-model/model.ts
pjanik: This is the most unlikely use case when we already have 10 plates, and 11th is added (I don't think that's possible in practice with current settings).

js/components/boundary-config-dialog.tsx
```diff
@@ -12,6 +12,7 @@ import BoundarySvg from "../../images/boundary.svg";
 import PlateArrow from "../../images/plate-arrow.svg";
 
 import css from "../../css-modules/boundary-config-dialog.less";
+import { plateHues } from "../plates-model/plate";
```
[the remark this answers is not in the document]
pjanik: Fixed.

PR #242, opened 2022-05-20
Eliminate React unique keys warning

kswenson:


pjanik (approved):
It was bothering me for a while. Thanks!

PR #246, opened 2022-05-23
Update map type button images

pjanik:
[#182081001]
[#182122398]
[#182122396]
[#182121588]

I've removed all non@3x versions, as we don't use them anyway.


kswenson (approved):
👍 LGTM

PR #247, opened 2022-05-23
Maintenance - update dependencies, React 17 -> 18, replace Enzyme tests with React Testing Lib

pjanik:
I had to replace Enzyme tests, as obviously there was no adaptor for Enzyme and workarounds that I found didn't work.


test/components/advanced-options.test.tsx
```diff
@@ -16,71 +17,55 @@ describe("AdvancedOptions component", () => {
...
+
+  it("respects `sidebar` config option and renders only chosen components - case 1", () => {
     config.sidebar = allOptions.map(e => e[0]);
-    let wrapper = mount(
-      <AdvancedOptions simulationStore={store} />
-    );
+    render(<AdvancedOptions simulationStore={store} />);
     allOptions.map(e => e[1]).forEach(label => {
-      expect(wrapper.text()).toEqual(expect.stringContaining(label));
+      expect(screen.queryByText(label)).toBeInTheDocument();
```
kswenson, line 37: For this usage (and others down below) I believe `getByText()` would be more conventional. `queryByText()` is generally used when zero or more than one instance are expected.

test/components/sortable-densities.test.tsx
```diff
@@ -18,16 +19,16 @@ describe("SortableDensities component", () => {
...
     config.densityWordInPlanetWizard = false;
 
-    wrapper = shallow(<SortableDensities simulationStore={store} />);
-    expect(wrapper.html()).toEqual(expect.stringContaining("BELOW"));
-    expect(wrapper.html()).toEqual(expect.stringContaining("ABOVE"));
-    expect(wrapper.html()).toEqual(expect.stringContaining("Click and drag to reorder the plates"));
+    render(<SortableDensities simulationStore={store} />);
+    expect(screen.queryByText("BELOW")).toBeInTheDocument();
+    expect(screen.queryByText("ABOVE")).toBeInTheDocument();
+    expect(screen.queryByText("Click and drag to reorder the plates")).toBeInTheDocument();
```
kswenson, line 32: As mentioned above, I think `getByText()` would be more conventional for all of these.

test/components/advanced-options.test.tsx
```diff
@@ -1,13 +1,14 @@
 import React from "react";
-import { mount } from "enzyme";
+import { render, fireEvent, screen } from "@testing-library/react";
+import "@testing-library/jest-dom";
```
kswenson, line 3: If you put this side-effect-only import into `setupTests.js` then you wouldn't have to import it separately in each test file.

kswenson (approved):

kswenson (approved):
👍 LGTM -- just a few suggestions about react-testing-library usage. I'm surprised that the update to React 18 was as simple as it was given the age of react-toolbox. I thought we might have to remove react-toolbox before updating.

pjanik:
> 👍 LGTM -- just a few suggestions about react-testing-library usage. I'm surprised that the update to React 18 was as simple as it was given the age of react-toolbox. I thought we might have to remove react-toolbox before updating.

It actually got broken in Cypress tests - I had to add a bunch of `ripple={false}` params, but that's not too bad. :)

test/components/sortable-densities.test.tsx
pjanik: Done.

test/components/advanced-options.test.tsx
pjanik: Done.

test/components/advanced-options.test.tsx
pjanik: Hm, this actually doesn't work for me. I've tried that before and it behaved strangely - sometimes one test would work, sometimes not. But when I ran `npm run test`, they consistently fail - there's an error about `toBeInTheDocument` not being defined.

PR #249, opened 2022-05-24
Fix tabs styles

pjanik:
Before:
Screenshot 2022-05-24 at 18 22 22
Now:
Screenshot 2022-05-24 at 18 22 02

It got broken a while ago, not sure when, so fixing it before I definitely leave the project. :)


kswenson (approved):
👍 LGTM

PR #255, opened 2022-07-18
Add fix for AWS error that failing S3 deploys

tejal-shah:


tejal-shah:
@pjanik , tectonic explorer deployments have been failing. This was also experienced in other repos a few weeks ago and I've just added the fix that Ethan proposed in those repos here.

pjanik:
@tejal-shah, perfect, thank you!

PR #259, opened 2022-09-02
Magma slug in convergent boundary should be measurable with temp/pressure tool

pjanik:
[[#182716184]](https://www.pivotaltracker.com/story/show/182716184)

Magma slug in convergent boundary isn't related to one field - it uses averaged position of multiple fields. That's why interaction doesn't return a single field object. But it returns rock type label. So, this PR updates the Measure Pressure/Temp tool to handle a case when only the rock label is available.


kswenson (approved):
👍 LGTM

PR #261, opened 2022-09-06
Fix black column at boundary during continental divergence

pjanik:
[[#182974865]](https://www.pivotaltracker.com/story/show/182974865)

The problem was caused by the divergent boundary field that didn't specify normalized age. So, late it was assumed to be 0 and this was causing `getRockCanvasPatternGivenNormalizedAge` function to return `#333` colo. This color should be only used for the oceanic divergent boundary as a base for the red overlay.


kswenson (approved):
👍 LGTM

PR #265, opened 2022-09-22
Reduce subduction magma slug width

pjanik:
The issue is well described in the PT story: https://www.pivotaltracker.com/story/show/182716121

This PR adds two params that will make the slug slightly smaller and make subduction plate subduct steeper (only visually) to avoid overlap of the subducting plate and the magma slug.

### DIFF
--- a/js/plates-view/render-cross-section.ts
+++ b/js/plates-view/render-cross-section.ts
@@ -666,7 +666,7 @@ class CrossSectionRenderer {
     const nativeWidth = 189;
     const nativeHeight = 70;
     const scale = 0.7;
-    const scaledWidth = scale * nativeWidth;
+    const scaledWidth = scale * nativeWidth * config.subductionMagmaSlugWidth;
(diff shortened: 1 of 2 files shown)


js/plates-view/render-cross-section.ts
```diff
@@ -679,6 +679,11 @@ class CrossSectionRenderer {
     ctx.ellipse(scaleX(avgLocation.x), scaleY(avgLocation.y), 0.45 * scaledWidth, 0.4 * scaledHeight,
       2 * Math.PI, 0, 2 * Math.PI);
 
+    // uncomment these lines to visualize/debug the hit-testing shape
+    // ctx.lineWidth = 1;
+    // ctx.strokeStyle = "black";
+    // ctx.stroke();
+
```
pjanik, line 686: I think I've accidentally removed this code before (it was yours and useful for debugging), so I'm re-adding it here.

kswenson (approved):
👍 Looks good

PR #267, opened 2022-09-27
Make sure that rocks deep under the surface have high grade metamorphism overlay

pjanik:
[#183128390]

Before fix:
Screenshot 2022-09-27 at 10 37 34
The area marked in red still has low and medium-grade metamorphism even though these rocks are deep under the surface and should be marked as high-grade metamorphic rocks. This PR is fixing it.

After fix:
Screenshot 2022-09-27 at 11 02 58

### DIFF
--- a/js/plates-view/render-cross-section.ts
+++ b/js/plates-view/render-cross-section.ts
@@ -280,7 +280,7 @@ class CrossSectionRenderer {
       }
       // Debug info, optional
       if (config.debugCrossSection) {
-        this.debugInfo(l1, b1, [i, `${f1.id} (${f1.plateId})`, x1.toFixed(1) + " km"]);
+        this.debugInfo(c1, l1, [i, `${f1.id} (${f1.plateId})`, x1.toFixed(1) + " km"]);
       }
       if (f1.magma && f1.magma.length > 0) {
         const isMagmaActive = this.drawMagma(f1.magma, f1, l1);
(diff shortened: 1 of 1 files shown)


js/plates-view/render-cross-section.ts
```diff
@@ -478,19 +478,26 @@ class CrossSectionRenderer {
...
-          const p1a = (new THREE.Vector2()).lerpVectors(p1, p4, METAMORPHISM_OROGENY_COLOR_STEP_0);
-          const p1b = (new THREE.Vector2()).lerpVectors(p1, p4, METAMORPHISM_OROGENY_COLOR_STEP_1);
-          // Divide vertical p2-p3 line into 3 sections (p2-p2a, p2a-p2b, p2b-p3).
-          const p2a = (new THREE.Vector2()).lerpVectors(p2, p3, METAMORPHISM_OROGENY_COLOR_STEP_0);
-          const p2b = (new THREE.Vector2()).lerpVectors(p2, p3, METAMORPHISM_OROGENY_COLOR_STEP_1);
-
-          this.fillMetamorphicOverlay("Low Grade Metamorphic Rock (Continental Collision)", field, p1, p2, p2a, p1a);
-          this.fillMetamorphicOverlay("Medium Grade Metamorphic Rock (Continental Collision)", field, p1a, p2a, p2b, p1b);
-          this.fillMetamorphicOverlay("High Grade Metamorphic Rock (Continental Collision)", field, p1b, p2b, p3, p4);
+          const kMetamorphismHighGradeOnlyThreshold = 0.015;
```
kswenson, line 481: I take it that you don't think it's worth elevating this to a `config` parameter?

kswenson (approved):
👍 Code looks good -- just one question about whether it's worth making a `config` param.

js/plates-view/render-cross-section.ts
pjanik: Yes, as this value is pretty much about filtering out the first field that is just subducting. So, there's not that much to configure and/or play with I guess.

PR #270, opened 2022-10-10
Fix earthquakes rendering

pjanik:
[[#183487545]](https://www.pivotaltracker.com/n/projects/2441249/stories/183487545)

This PR fixes an issue most likely caused by the THREE.JS version upgrade. ALPHATEST uniform is no longer available. I've tried to use `alphaTest` that seemed to be the right name based on the source code, but it still didn't work. It's not really critical, so I just hardcoded 0.5 (we could also not discard these pixels and nothing bad would happen).


kswenson (approved):
👍 LGTM

PR #271, opened 2022-10-11
Add divergentBoundary2 preset

pjanik:
[#183500738]


pjanik:
Going to merge it without review, as it's just a new preset requested by Chris Lore.

PR #274, opened 2022-11-18
Replace plate merging with plate grouping

pjanik:
[#183678655]

Plate merging used to happen when two adjacent plates had a relative speed close to 0. However, the project team didn't like the process of "merging" when two plates were essentially replaced by a single one.

I've added plate grouping as an alternative. Plates are not merged, they'll still be two separate entities. However, the group will "glue" them together and they'll behave like one rigid body/plate. This is mainly achieved by calculating the moment of inertia and angular acceleration. 

I've also updated some simulation steps to make sure that we don't recalculate the moment of inertia or angular acceleration unnecessarily, as both operations are very CPU expensive (they iterate over all the fields).


dougmartin (approved):
Looks good :+1:

lublagg (approved):
Looks good to me 👍

PR #276, opened 2022-11-22
183681490 popup

pjanik:
https://www.pivotaltracker.com/story/show/183681490

I think the PT story description provides a good context.

It's an example of a change where I had to modify the code ran in the web worker and then send the new property to the UI thread. See model-store.ts and `relativeMotionStopped`. If that's not very clear, we could meet tomorrow and we can review that in 5-10 minutes on Zoom. Another option would be to send a postMessage when the model is stopped, but I thought that using a property might be easier/cleaner. I'd probably try to limit the number of postMessages traveling back and forth, as it's more difficult to follow for me.

The new dialog is based on `DraggableDialog`. I've extracted it from the existing dialog used to configure the boundary type.

### DIFF
--- a/css-modules/boundary-config-dialog.less
+++ b/css-modules/boundary-config-dialog.less
@@ -1,37 +1,6 @@
 .boundaryConfigDialog {
-
-  :global(.MuiDialogTitle-root) {
-    font-size: 14px;
-    font-weight: bold;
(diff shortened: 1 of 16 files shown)


dougmartin (approved):
Looks good :+1:

PR #277, opened 2022-11-24
Update styles of earthquakes caveat notice to avoid conflicts with cross section buttons

pjanik:
[[#183763407]](https://www.pivotaltracker.com/story/show/183763407)

I think the issue was caused by the caveat notice:
issue
Even if it's invisible, it's container was in the DOM. So, the main fix was to add `pointer-events: none` not to block the buttons that it overlaps.

Also, I've slightly updated styles to make it less likely to overlap the cross-section. Before:
before
After:
after


dougmartin (approved):
Looks good :+1:

PR #279, opened 2022-11-25
Make sediments rendering opional

pjanik:
[[#183763592]](https://www.pivotaltracker.com/story/show/183763592)

This PR adds a new "Sediments" checkbox in the advanced options + a new URL param `sediments` that can be used to set the initial value of this checkbox (=> and param in the `SimulationStore`).

This new param can be used to show/hide all the sediments in the model, both globe and cross-section view. It replaces the old `sedimentsInPlanetView` URL param that was less capable and not modifiable by users in the runtime.

Since this new param is "dynamic" (=> can be changed by the user, not only via the URL config), the implementation had to touch much more classes. Note that I also have to recalculate the field `elevation`. When we remove sediments, we need to consider that. Otherwise, it'd mess up rock layers rendering in the cross-section. Another side effect is that you can see topography being updated when you turn on and off the` sediments` prop. But I think it actually makes sense.

I've also discovered MobX `reaction` that I think it's much cleaner than magic `autorun`. I've tried to refactor most of the code, but it becomes pretty wordy and they're not fully equivalent - `autorun` calls its callback initially while `reaction` does not. So, this caused some subtle bugs and I was worried I wouldn't catch all of them. But if you ever need to use it, I'd recommend reading about differences and I'd probably lean towards `reaction` because of the clarity.


js/plates-model/crust.ts
```diff
@@ -104,10 +104,17 @@ export default class Crust {
     return result;
   }
 
-  get topRockType() {
-    if (!config.sedimentsInPlanetView) {
+  get sedimentThickness() {
+    if (isSediment(this.rockLayers[0].rock)) {
```
dougmartin, line 108: Maybe document why you are only checking index 0 here?  The other getters seem to check all the rock layers.

dougmartin (approved):
Looks good :+1:  Just had one documentation comment but maybe that is self evident when you follow the thread of the sediment code or it is documented elsewhere.

kswenson:
> I've also discovered MobX `reaction` that I think it's much cleaner than magic `autorun`. I've tried to refactor most of the code, but it becomes pretty wordy and they're not fully equivalent - `autorun` calls its callback initially while `reaction` does not. So, this caused some subtle bugs and I was worried I wouldn't catch all of them. But if you even need to use it, I'd recommend reading about differences and I'd probably lean towards `reaction` because of the clarity.

FYI, `reaction` takes an `options` argument which allows you to specify `firstTime: true` so that you _can_ have a `reaction` behave like `autorun` in that respect if necessary/desired.

js/plates-model/crust.ts
pjanik: Done in the recently pushed commits.

PR #281, opened 2022-11-30
Update model to avoid low elevation areas trapped between mountains

pjanik:
[[#183128190]](https://www.pivotaltracker.com/story/show/183128190)

PT story provides a good description of the issue. I've modified a few separate parts of the model to avoid the steep slopes around the convergent boundary that was creating these deep valleys, often lower than the sea level (hence the blue color and impression of "water" being trapped there).

You can compare the new and old behavior:
- [New version](https://tectonic-explorer.concord.org/branch/continental-collision/index.html?modelId=dcd796fb-9c59-428b-9421-0aa77431b7c3&rocks=true)
- [Old version](https://tectonic-explorer.concord.org/branch/master/index.html?modelId=dcd796fb-9c59-428b-9421-0aa77431b7c3&rocks=true)

There's no water in the updated version, although there's still some valley following the boundary. If you look at the cross-section, you can also see some differences. Shale, limestone, and sandstone are being accumulated around the edge instead of being pushed underground.

Amy seems to like the new behavior, so I'm opening the PR.

### DIFF
--- a/js/plates-model/crust.ts
+++ b/js/plates-model/crust.ts
@@ -369,9 +369,10 @@ export default class Crust {
     }
   }
 
-  // Used during subduction of the oceanic plate (subduction) or when continental plate goes
-  // under another continental plate (folding).
-  subductOrFold(timestep: number, neighboringCrust: Crust[], relativeVelocity?: THREE.Vector3) {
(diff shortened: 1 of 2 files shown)


dougmartin (approved):
Looks good :+1:

PR #282, opened 2022-12-01
Divide plates using age value, rm plate merging code

pjanik:
[[#183681517]](https://www.pivotaltracker.com/story/show/183681517)

Project team requested very new behavior/rules that define plate division. It used to be based on the plate size and the new boundary was very random.

Now, the division happens when the freshly added crust becomes old enough (=> heavy). In this case, the plate breaks into two plates, creating a new convergent boundary defined by age differences. The PT story provides lots of context and description of expected behavior. I've added a few other rules that are not defined in PT story to avoid creating multiple small plates. 

I've removed the old division and merge code (merging wasn't used for a while, since I added grouping: https://github.com/concord-consortium/tectonic-explorer/pull/274). That's why I could completely remove one of the test files.


dougmartin (approved):
Looks good :+1:

pjanik:
I accidentally merged this PR into the wrong branch, so I merged `183681517-plate-division-by-age` into `master` manually.

PR #283, opened 2022-12-02
Handle preset case in the relative motion stopped dialog

pjanik:
[#183681490]

A small change to address the issue found by Saravanan:
https://www.pivotaltracker.com/story/show/183681490/comments/234509311


dougmartin (approved):
Looks good :+1:

PR #286, opened 2022-12-08
Add data collection draft UI, disabled by default

pjanik:
[#183946366]

This PR adds a new button named "Collect Data". It can be used to pick a point in the cross-section. UI is in a very draft state, we don't have UI mocks/specs yet. But this UI will be disabled by default, so I think it make sense to keep merging updates rather than keep all the work in a branch.

Demo:
https://tectonic-explorer.concord.org/branch/data-collection/index.html?rocks=true&preset=subduction&showCollectDataButton=true


dougmartin (approved):
Looks good :+1:

PR #287, opened 2022-12-12
Manage LARA interactive state with dataset

pjanik:
[[#183946419]](https://www.pivotaltracker.com/story/show/183946419)

This PR adds LARA Interactive API support. Tectonic Explorer can now save and load interactive state that mostly includes dataset that is going to be rendered by a new table interactive.

### DIFF
--- a/js/stores/simulation-store.ts
+++ b/js/stores/simulation-store.ts
@@ -2,6 +2,7 @@ import { observable, computed, action, runInAction, autorun, makeObservable } fr
 import config, { Colormap } from "../config";
 import * as THREE from "three";
 import isEqual from "lodash/isEqual";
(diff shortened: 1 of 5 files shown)


js/stores/simulation-store.ts
```diff
@@ -676,23 +694,48 @@ export class SimulationStore {
...
+    this.saveInteractiveState();
   }
 
   // Helpers.
+  saveInteractiveState() {
+    setInteractiveState<IInteractiveState>({
+      answerType: "interactive_state",
+      dataset: {
+        type: "dataset",
+        version: 1,
```
dougmartin, line 728: I like the future-proof versioning.

dougmartin (approved):
Looks good :+1:

PR #288, opened 2022-12-12
Render data samples in the cross-section view

pjanik:
[[#184003233]](https://www.pivotaltracker.com/story/show/184003233)

The basic cross-section markers rendering. They are going to have different shapes and need to be interactive most likely, so this is just a temporal solution.
Screenshot 2022-12-12 at 19 34 48


dougmartin (approved):
Looks good :+1:

PR #290, opened 2022-12-22
Save globe view and cross-section snapshots in interactive state

pjanik:
[#184027345]

Tectonic Explorer will now save two screenshots when it saves the interactive state. It isn't visible for users (until they look at the TecRock Table), and I'm trying to ensure that we never overwrite newer screenshots with older ones. This could happen easily, as the Shutterbug response time is very variable and usually the first request is the slowest.

### DIFF
--- a/js/shutterbug-support.ts
+++ b/js/shutterbug-support.ts
@@ -1,4 +1,6 @@
 import Shutterbug from "shutterbug";
+import { CROSS_SECTION_CANVAS_ID } from "./plates-view/cross-section-3d";
+import { PLANET_VIEW_CANVAS_ID } from "./plates-view/planet-view";
 
 export function enableShutterbug(appClassName: string) {
   Shutterbug.enable("." + appClassName);
@@ -19,3 +21,37 @@ function beforeSnapshotHandler() {
     }
   });
 }
+
+export function takeSnapshot(selectorOrDomElement: string | HTMLElement) {
+  return new Promise((resolve, reject) => {
+    Shutterbug.snapshot({
+      selector: selectorOrDomElement,
+      done: (snapshotUrl: string) => {
+        resolve(snapshotUrl);
+      },
+      fail: (jqXHR: any, textStatus: string, errorThrown: any) => {
+        console.error("Shutterbug request failed", textStatus, errorThrown);
+        reject(errorThrown);
+      }
+    });
+  });
+}
+
+export function takePlanetViewSnapshot() {
+  return takeSnapshot(`#${PLANET_VIEW_CANVAS_ID}`);
+}
+
+export function takeCrossSectionSnapshot() {
+  const canvas = document.getElementById(CROSS_SECTION_CANVAS_ID);
+  let oldBgColor = "";
+  if (canvas) {
+    oldBgColor = canvas.style.backgroundColor;
+    canvas.style.backgroundColor = "black";
+  }
+  const promise = takeSnapshot(`#${CROSS_SECTION_CANVAS_ID}`);
+  if (canvas) {
+    canvas.style.backgroundColor = oldBgColor;
+  }
+  return promise;
+}
+
--- a/js/stores/simulation-store.ts
+++ b/js/stores/simulation-store.ts
@@ -2,7 +2,7 @@ import { observable, computed, action, runInAction, autorun, makeObservable } fr
 import config, { Colormap } from "../config";
 import * as THREE from "three";
 import isEqual from "lodash/isEqual";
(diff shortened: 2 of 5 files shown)


js/shutterbug-support.ts
```diff
@@ -19,3 +21,37 @@ function beforeSnapshotHandler() {
...
+      fail: (jqXHR: any, textStatus: string, errorThrown: any) => {
+        console.error("Shutterbug request failed", textStatus, errorThrown);
+        reject(errorThrown);
+      }
+    });
+  });
+}
+
+export function takePlanetViewSnapshot() {
+  return takeSnapshot(`#${PLANET_VIEW_CANVAS_ID}`);
```
dougmartin, line 41: Just curious why you interpolate this since it is already a string?

js/shutterbug-support.ts
```diff
@@ -19,3 +21,37 @@ function beforeSnapshotHandler() {
...
+}
+
+export function takeCrossSectionSnapshot() {
+  const canvas = document.getElementById(CROSS_SECTION_CANVAS_ID);
+  let oldBgColor = "";
+  if (canvas) {
+    oldBgColor = canvas.style.backgroundColor;
+    canvas.style.backgroundColor = "black";
+  }
+  const promise = takeSnapshot(`#${CROSS_SECTION_CANVAS_ID}`);
```
dougmartin, line 51: I know we don't want to wait until the full snapshot url is resolved but isn't there a risk that the background color is reset before the snapshot is taken?

js/stores/simulation-store.ts
```diff
@@ -734,20 +737,53 @@ export class SimulationStore {
...
+      // This is client side timestamp and it's possible that the clock is set incorrectly. However, this is perfectly
+      // fine here, as it's only used to compare the order of requests.
+      const requestTimestamp = Date.now();
+      const snapshotPromises = [takePlanetViewSnapshot()];
+      if (this.crossSectionVisible) {
+        snapshotPromises.push(takeCrossSectionSnapshot());
       }
-    });
+      Promise.all(snapshotPromises).then(([planetViewSnapshot, crossSectionSnapshot]: string[]) => {
+        if (requestTimestamp < this.lastSnapshotRequestTimestamp) {
```
dougmartin, line 774: Since `takeSnapshot` can reject I think we need a `.catch()` here that just sets `newState` and skips the snapshots.

dougmartin (changes requested):
Looks good - I think we do need a catch in `saveInteractiveState` in case the snapshots fail.

js/shutterbug-support.ts
pjanik: There's `#` in front of it. I guess it melts with `$`. So, it's `"#" + PLANET_VIEW_CANVAS_ID`.

js/shutterbug-support.ts
pjanik: Fair question. I believe there's no risk, but mostly because I remember how Shutterbug JS lib works. :wink:
If there are no iframes, it'll clone and process HTML elements synchronously.
If there are iframes, it'll wait for their response, so even client-side processing will be async. And that would be the problem that you brought up.

We're dealing with canvas only, so the client-side processing will be fully synchronous.

There's no way to react to client-side processing only. If we wanted to be 100% sure, we'd need to wait for this promise to resolve. But that happens only after the server responds, so multiple seconds and it'll be visible for users.

I think it'd leave it this way, as even if we restore transparent background too soon, nothing bad will happen. Cross-section screenshot will just have a white background in the table. Black looks better, but the white one is fully functional. This juggling with the background is a minor aesthetic tweak.

js/stores/simulation-store.ts
pjanik: Note that newState is set even before we try to take snapshots. Line 760 above. So, in most cases, we'll update the state twice, but that minimizes the risk of losing data. Sample data will be saved immediately, and we'll wait only for snapshots. That's why we don't really need to `catch` in this case.

Also, next PR adds `.catch` anyway:
https://github.com/concord-consortium/tectonic-explorer/pull/291

pjanik:
@dougmartin, thanks for catching these things. I've replied in the inline comments. I guess there might be nothing to change in fact, just tried to explain better what happens, so I re-requested the review.

dougmartin (approved):
Looks good :+1:

js/shutterbug-support.ts
dougmartin: Oops. I missed the `#` - makes sense now.

PR #295, opened 2022-12-28
Highlight the most recent pin placed when collect data dialog is visible

pjanik:
[#184085688]

Also, two related assets were updated as Kiley re-exported them.


kswenson (approved):
👍 Looks good

PR #298, opened 2023-01-09
Fix issues related to data saving

pjanik:
[#184074513]
[#184074641]

This PR disables initial interactive state saving, as that's not really useful and it was causing some issues.
Also, the dialog informing about data saving won't be shown if there are samples to save.
Finally, the forward navigation is re-enabled with a small delay, so it's less likely user is able to change page when AP is saving the updated state in Firestore. Plus I've added one MobX `action` wrapper to avoid warnings.

### DIFF
--- a/js/stores/simulation-store.ts
+++ b/js/stores/simulation-store.ts
@@ -825,13 +825,15 @@ export class SimulationStore {
             console.error(error);
           }
         })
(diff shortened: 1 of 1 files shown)


dougmartin (approved):
Looks good :+1:

PR #299, opened 2023-01-09
Fix navigation unlocking

pjanik:
[#184074641]

A silly bug, `setNavigation` call was outside the `if` clause.


dougmartin (approved):
Looks good :+1:

PR #300, opened 2023-01-10
Add `js/shared` directory with shared helpers and assets

pjanik:
[[#184027461]](https://www.pivotaltracker.com/story/show/184027461)

This PR might be difficult to review because of the number of assets and changes.

The main goal of this PR was to restyle the data collection dialog by adding the table specified here:
https://app.zeplin.io/project/60c9c0d5060353bd2bb10172/screen/6393a850ba25a3b78759acb3

However, almost identical table will need to be rendered by the TecRock table interactive that lives in the `question-interactives` repo. So, my idea was to create a `shared` package with all the useful JS helpers (function to generate table cells) and assets (plenty of SVG and PNG images). This package is currently defined in `js/shared`. I've updated `package.json` and `webpack.config.js` so this library can be built and accessible.

Other changes:
- Table columns can be now defined by a new URL param: `dataSampleColumns`.
- I had to split multiple SVG diagrams into two separate SVGs (eg Granite diagram into Granite Iron Content and Granite Cooling). 
- I've updated Rock Key data structures to re-use the new SVGs, patterns, and all the info defined in `js/shared`, so there's no code duplication and assets will stay in sync.
- I've also cleaned up and slightly reorganized other assets, as I was in the right context.

I've tried to NPM link this new shared library and use it in TecRock Table question-interactives - it seemed to work fine. 
PNGs are very small (20x20px) so they're inlined in JS bundle. SVGs are handled by Webpack SVGr and also inlined into JS bundle. This saves us a lot of headaches I think.

However, I couldn't get types working so far (but I haven't tried too much). @dougmartin, @kswenson, any chances you're familiar with a nice pattern that lets us generate TypeScript type definitions using Webpack? It'd be nice if it also respected entry points. I've mostly seen patterns that use `tsc` directly which wouldn't work well with Webpack entry points I think.

Also, somehow related - any thoughts on how to define a NPM library that needs to export its own, large image assets? It's not the case here, as luckily we can inline all the assets in the JS bundle. But I wonder how we approach it, and how we later load them on the other side (library consumer).


dougmartin (approved):
Looks good :+1: 

I filtered out the asset files and just really looked at the source code changes along with the build file changes.  Makes sense to me.

PR #301, opened 2023-01-11
Convert to monorepo, add TecRock Table as a new package, extract Shared to package

pjanik:
[[#184185704]](https://www.pivotaltracker.com/story/show/184185704)

The goal is to update TecRock Table rendering, so it uses the same helpers and assets as the data collection dialog in Tectonic Explorer simulation.

I've decided to convert this repository to monorepo and move TecRock Table from the question-interactives repo: https://github.com/concord-consortium/question-interactives/tree/master/packages/tecrock-table.

That partially reverts some of the previous updates where `TecRock Shared` was defined as a new package. But it'd be annoying to keep updating it long-term, plus there's zero value of this package for other projects.

I've started a Slack discussion about it: 
https://concord-consortium.slack.com/archives/C0303QUNU/p1673447096770299:

> I have a question about architecture of our repositories and monorepos. TecRock has currently following projects that need to interact and depend on each other:
Tectonic Explorer - the main app/simulation
TecRock Table - interactive that renders table
TecRock Helpers - assets and rendering helpers that are used both by Tectonic Explorer and TecRock Table. We don't want to keep two copies of these assets in each repo, as there's a lot of them, and they need to stay in sync.
Question Interactives Helpers - LARA Interactive API helpers, hooks, and react-json-schema-form helpers. All the utilities that let you create a new interactive with authoring interface easier. These already exist as NPM package, as I did it during refactoring of the question-interactives into monorepo.

> We have two repositories now where these projects live:
> - tectonic-explorer repo (original repo of the simulation)
> - question-interactives repo

> And I can see a few options how to organize these projects:
> 1. tectonic-explorer repo contains simulation Tectonic Explorer and TecRock Helpers that are pushed to NPM. question-interactives contains its helpers obviously and TecRock Table that needs to install TecRock Helpers from NPM.
> 2. tectonic-explorer repo becomes a monorepo with 3 packages: Tectonic Explorer , TecRock Table and TecRock Helpers.  It makes using shared TecRock helpers easy, and we don't need to push them to NPM each time we want to update some asset there. There's zero chances some external, non-TecRock project will be interested in these helpers. TecRock Table needs to install Question Interactives Helpers from NPM, but these helpers might be used by many other interactives, as they're generic. So it's perfectly justified to keep them in NPM registry.

> I managed to have both options up and running, so there's a comfort of choosing one that we like more. :wink:
Pros of 1: TecRock Table  can take advantage of question-interactives infrastructure, Cypress testing patterns, deployment. We slowly gravitate into having all interactives in one monorepo.
Cons of 1: Working on Tectonic Explorer simulation and TecRock table might get annoying, as it'll require constant NPM package updates (TecRock Helpers). It can be risky to put too many projects in one monorepo in case we're "not ready for it" yet (maybe our patterns are not tested well enough etc., and it can become a trap long-term).

> Pros of 2: TecRock-related projects live in one monorepo. Usually we work on them together, and that's way simpler, as updates to shared assets don't include pushing and pulling from NPM (or using npm link ). We're forced to slowly improve Question Interactives Helpers
package.
Cons of 2: Our interactives are spread between various repos. We're not becoming second Google with one repo for everything. But it might be safer approach while we test monorepo pattern more. We can't easily reuse Cypress testing patterns from question-interactives repo.

This PR essentially implements 2 approach.


dougmartin (approved):
Looks good :+1:

PR #302, opened 2023-01-11
TecRock Table updates

pjanik:
[[#184185704]](https://www.pivotaltracker.com/story/show/184185704)

Two main changes in this PR:
1. I've given up on using `IDataset` in Tectonic Explorer interactive state. It became pointless, as this dataset cannot be rendered by any generic interactive. Most of the data in the table are derived from rock type, and there are plenty of SVGs that need to be injected. Also, now TecRock Table and Tectonic Explorer live in the same repository and use `shared` package, so there's no problem in sharing the interactive state interface between them.
2. I've updated TecRock Table rendering to match table in the Tectonic Explorer Data Collection dialog.

### DIFF
--- a/packages/tecrock-table/src/components/runtime.tsx
+++ b/packages/tecrock-table/src/components/runtime.tsx
@@ -1,22 +1,20 @@
 import React, { useEffect } from "react";
 import classNames from "classnames";
 import { IRuntimeQuestionComponentProps } from "@concord-consortium/question-interactives-helpers/src/components/base-question-app";
-import { IAuthoredState, ITectonicExplorerInteractiveState } from "../types";
+import { IAuthoredState } from "../types";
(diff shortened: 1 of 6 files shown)


packages/tecrock-table/src/components/runtime.tsx
```diff
@@ -57,20 +55,28 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
...
-            <tbody>
+        <table>
+          <tbody>
+            <tr>
               {
-                dataset?.rows.map(row => (
-                  <tr key={row[0]}>
-                    { row.slice(1).map((value, idx) => (<td key={idx}>{ value }</td>)) }
-                  </tr>
+                dataSampleColumns && dataSampleColumns.map((column: DataSampleColumnName) => (
```
dougmartin, line 62: Just curious why you moved it out of `<thead>`.

dougmartin (approved):
Looks good :+1:

PR #303, opened 2023-01-11
Update authoring page, let authors configure dataSampleColumns

pjanik:
[[#184185532]](https://www.pivotaltracker.com/story/show/184185532)

Small PR that updated authoring page.

### DIFF
--- a/packages/tectonic-explorer/js/components/authoring.tsx
+++ b/packages/tectonic-explorer/js/components/authoring.tsx
@@ -10,6 +10,7 @@ import config, { Colormap } from "../config";
 import { INTERACTION_LABELS } from "./interaction-selector";
 import { STEPS_DATA } from "./planet-wizard";
 import presets from "../presets";
+import { dataSampleColumnLabel } from "@concord-consortium/tecrock-shared";
 
 import css from "../../css-modules/authoring.less";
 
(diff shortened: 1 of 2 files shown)


packages/tectonic-explorer/js/components/authoring.tsx
```diff
@@ -99,6 +102,7 @@ const AUTOCOMPLETE_OPTIONS: Record<string, any> = {
   sidebar: {
     "interactions": "Interactions",
     "metamorphism": "Metamorphism",
+    "sediments": "Sediments",
```
dougmartin, line 105: This seems a little unrelated to the description of the change.  Is this needed?

dougmartin (approved):
Looks good :+1: 

I just had a question about adding `sediments` to the `sidebar` object, in case that was not intended.

packages/tectonic-explorer/js/components/authoring.tsx
pjanik: Right, I didn't mention that in the description, but it was intended. It should have been added before when I actually implemented the Sediments checkbox. It was impossible to configure it on the authoring page, and it was also breaking some logic that was trying to compare the current config value with defaults and update the URL if necessary. So, the authoring page was adding a sidebar option to the URL even when the author didn't modify defaults.

PR #304, opened 2023-01-12
Add notes support

pjanik:
[#184214544]

This PR ensures that notes texarea (TE dialog) and notes column (TecRock Table) are disabled when "notes" value is not present in `dataSampleColumns` config value. I've also fixed some styles and ensured that column are always rendered in the same order.


dougmartin (approved):
Looks good :+1:

PR #306, opened 2023-01-12
Fix a bug in Data Collection dialog (notes textarea)

pjanik:
The bug was pretty weird, but it's well described by Saravanan here:
https://www.pivotaltracker.com/story/show/184074624/comments/235007860

Half of the problem was:
```
this.currentDataSample.notes = notes;
```
instead of:
```
this.currentDataSample = {
  ...this.currentDataSample,
  notes
};
```

but that revealed another problem in the dialog. The notes textarea was losing focus on each key stroke. That's fixed by memoization of the PaperComponent.

### DIFF
--- a/packages/tectonic-explorer/js/components/data-collection-dialog.tsx
+++ b/packages/tectonic-explorer/js/components/data-collection-dialog.tsx
@@ -15,11 +15,12 @@ interface IProps {
 }
 
 // patterned after https://mui.com/components/dialogs/#draggable-dialog
-export const DataCollectionDialog = ({ currentDataSample, onClose, onSubmit, onNotesChange }: IProps) => {
+export const DataCollectionDialog: React.FC<IProps> = ({ currentDataSample, onClose, onSubmit, onNotesChange }) => {
 
   const handleNotesChange = useCallback((event: React.ChangeEvent<HTMLTextAreaElement>) => {
     onNotesChange(event.target.value);
   }, [onNotesChange]);
+
 
   const rockRowData = dataSampleToTableRow(currentDataSample);
 
@@ -59,7 +60,7 @@ export const DataCollectionDialog = ({ currentDataSample, onClose, onSubmit, onN
         </table>
         {
           notesEnabled &&
-          <textarea key="notes" placeholder="Add notes…" value={currentDataSample.notes} onChange={handleNotesChange} />
+          <textarea placeholder="Add notes…" value={currentDataSample.notes} onChange={handleNotesChange} />
         }
       </div>
       <DialogActions>
--- a/packages/tectonic-explorer/js/components/draggable-dialog.tsx
+++ b/packages/tectonic-explorer/js/components/draggable-dialog.tsx
@@ -1,4 +1,4 @@
-import * as React from "react";
+import React, { useMemo } from "react";
 import classNames from "classnames";
 import Draggable from "react-draggable";
 import Dialog from "@mui/material/Dialog";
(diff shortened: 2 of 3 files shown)


packages/tectonic-explorer/js/components/data-collection-dialog.tsx
```diff
@@ -59,7 +60,7 @@ export const DataCollectionDialog = ({ currentDataSample, onClose, onSubmit, onN
         </table>
         {
           notesEnabled &&
-          <textarea key="notes" placeholder="Add notes…" value={currentDataSample.notes} onChange={handleNotesChange} />
+          <textarea placeholder="Add notes…" value={currentDataSample.notes} onChange={handleNotesChange} />
```
dougmartin, line 63: With the addition (in the other PR) of the `id` field, would a key of "notes_" + currentDataSample.id be good to add here or is this dialog modal so that `currentDataSample` doesn't change while it is displayed?

packages/tectonic-explorer/js/stores/simulation-store.ts
```diff
@@ -769,7 +772,7 @@ export class SimulationStore {
       version: 1,
       answerType: "interactive_state",
       dataSampleColumns: config.dataSampleColumns,
-      dataSamples: toJS(this.dataSamples) // copy MobX observable array to plain JS array
+      dataSamples: JSON.parse(JSON.stringify(this.dataSamples)) // copy MobX observable array to plain JS array
```
dougmartin, line 775: Just curious why you changed it to this.  Does `toJS` not do deep copies?

packages/tectonic-explorer/js/stores/simulation-store.ts
```diff
@@ -820,6 +823,7 @@ export class SimulationStore {
           // Mark the request as processed.
           if (requestTimestamp === this.lastSnapshotRequestTimestamp) {
             this.lastSnapshotRequestTimestamp = null;
+            // this.playing = true;
```
dougmartin, line 826: Did you want to leave this in?

dougmartin (approved):
Looks good :+1: 

Just some inline comments with questions that don't block merging.

packages/tectonic-explorer/js/stores/simulation-store.ts
pjanik: I have committed that accidentally. It was part of my debugging and I was testing if it wasn't a problem. `toJS` is enough, I'll update it in the future comments. Thanks.

packages/tectonic-explorer/js/stores/simulation-store.ts
pjanik: Nope, I'll remove it, thanks.

pjanik:
> Just some inline comments with questions that don't block merging.

Thanks, I've addressed most of them, so I'm merging the PR.

packages/tectonic-explorer/js/components/draggable-dialog.tsx
```diff
@@ -37,7 +37,9 @@ interface IProps {
 
 // patterned after https://mui.com/components/dialogs/#draggable-dialog
 export const DraggableDialog: React.FC<IProps> = ({ backdrop, onClose, title, offset, initialPosition, children }) => {
-  function PaperComponent(props: PaperProps) {
+  // PaperComponent needs to be memoized, as otherwise it's recreated on every render. This might lead to unexpected
+  // side effects like losing input focus on each render.
+  const PaperComponentMemoized = useMemo(() => function PaperComponent(props: PaperProps) {
```
kswenson, line 42: You could also move the definition of `PaperComponent` outside `DraggableDialog` altogether and pass in the `offset` as a prop (which I think is the only additional context required) which would avoid the re-creation of the component on every render (which is still occurring--the `useMemo()` is just preventing downstream code from noticing 😉) and eliminate the need for `useMemo()`.

pjanik: I was hoping I could do that, but note that `PaperComponent` is later passed to another component as property with constructor/class value, not even JSX element. I cannot modify its props and add `offset` there, as I don't have control over what its parent will pass to it. It'd need to be something like HOC, but we'd be back to the same problem I think.

PR #307, opened 2023-01-12
Clear data samples immediately after user confirms that he wants to restart data collection

pjanik:
[#184074513]

This PR addresses the comment made by Saravanan and then confirmed by Trudi:
https://www.pivotaltracker.com/story/show/184074513/comments/235006523

Data should be cleared immediately after user confirms that they want to restart data collection.


dougmartin (approved):
Looks good :+1:

PR #308, opened 2023-01-13
Prevent cross section rotation in data collection mode

pjanik:
[[#184186030]](https://www.pivotaltracker.com/story/show/184186030)

This PR:
1. Locks CS view in the data collection mode (first commit)
2. Fix a small random bug - sometimes data collection dialog could be visible even after user extits the data collection mode (second commit)


dougmartin (approved):
Looks good :+1:

PR #310, opened 2023-01-13
Update dialog buttons

pjanik:
[[#184027461]](https://www.pivotaltracker.com/story/show/184027461)

This PR updates buttons in the data collection dialog and makes them reusable.
UI Specs: https://app.zeplin.io/project/60c9c0d5060353bd2bb10172/screen/6393a850ba25a3b78759acb3

It also fixes this bug:
https://www.pivotaltracker.com/story/show/184224980
It was caused by postcss config not being recognized. I have no idea why it has to be in the top-level directory, but I didn't have time to debug it. Long-term, we'll try to migrate away from dead react-toolbox library anyway.


packages/tectonic-explorer/js/components/data-collection-dialog.tsx
```diff
@@ -26,7 +27,7 @@ export const DataCollectionDialog: React.FC<IProps> = ({ currentDataSample, onCl
 
   const notesEnabled = config.dataSampleColumns.includes("notes");
   // When user is collecting data, notes should show as a separate textfield under the table,
-  // so notes column require special handling. Also, there's no need to display pin id / number.
+  // so notes column require special handling. Also, there"s no need to display pin id / number.
```
dougmartin, line 30: The single quote to double quote looks like a search/replace mistake :)

dougmartin (approved):
Looks good :-1:

packages/tectonic-explorer/js/components/data-collection-dialog.tsx
pjanik: It is. :) Thanks, I'll fix it later, so I don't need to jump between branches now.

pjanik: It is. :) I'll fix it later, so I don't need to jump between branches now.

PR #312, opened 2023-01-15
TecRock table: let users expand snapshots in a lightbox

pjanik:
[[#184186001]](https://www.pivotaltracker.com/story/show/184186001)

This PR lets user expand snapshots in a lightbox:
Screenshot 2023-01-15 at 18 28 19

There are two separate buttons, as the current Interactive API `showModal({ type: "lightbox" })` accepts only one image URL. 

I could use  `showModal({ type: "dialog" })` and render an iframe with two images together, but this approach has some issues. AP dialog implementation is slightly broken and doesn't show the close icon and doesn't let the user close the dialog using backdrop (I've added a PT bug: https://www.pivotaltracker.com/story/show/184232312). Also, the dialog size won't match the image and there's not much we could do about it. It'd be also more code here. So, I'm trying this approach for now, as it's cleaner and lightbox UX is nicer for users.

### DIFF
--- a/packages/tecrock-table/src/assets/zoom-in.svg
+++ b/packages/tecrock-table/src/assets/zoom-in.svg
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<svg viewBox="0 0 200 200" xmlns="http://www.w3.org/2000/svg">
+  <g>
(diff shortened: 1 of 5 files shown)


dougmartin (approved):
Looks good :+1:

PR #313, opened 2023-01-15
Workaround issue caused by the fact that 3D cross-section is not a cube, but trapezoid

pjanik:
This PR fixes / workarounds issue described here: [[#184137645]](https://www.pivotaltracker.com/story/show/184137645)

The problem is quite tricky. The 3D cross-section code was making an assumption that front and back wall have the same dimensions/width. This is not true due to the nature of the spherical globe surface. Often, back wall is shorter when measured in kilometers, but it's rendered with the same width.

That leads to issues observed by Saravanen. It also affects the rock picker tool which will show a different rock that you're pointing when you try to pick something on the right side of the back cross-section wall.

The proper fix would be to use custom buffer geometries for all the cross-section walls and construct a 3D trapezoid. But it didn't seem realistic with the time I had left. So, instead, I've used the existing code but just made sure that the dimensions are now correct. 

It ensures that data and rock sampling are correct on the back wall, but we might see small gaps between walls:
cs-gaps
But this is only a visual issue, so I think it's better than the previous state which was a functional problem, potentially frustrating for students.

I'll add a new PT bug to address this new, visual issue later.

### DIFF
--- a/packages/tectonic-explorer/js/plates-view/cross-section-3d.ts
+++ b/packages/tectonic-explorer/js/plates-view/cross-section-3d.ts
@@ -217,41 +217,50 @@ export default class CrossSection3D {
     renderCrossSection(this.leftWallCanvas, data.dataLeft, dataSamples.left, options);
     this.leftWallTexture.needsUpdate = true;
 
-    const width = this.frontWallCanvas.width;
+    // Note that front wall width is slightly different than the back wall, as cross-section is drawn on a spherical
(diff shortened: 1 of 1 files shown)


pjanik:
@dougmartin, I'll go ahead and merge this PR, since you've reviewed all the other PRs and this one would be blocking me from `master` branch for the demo. 

If you have some comments, please still add them here after the merge, and I'll address them.

PR #314, opened 2023-01-15
Replace native alert with custom dialog that asks for clearing previously saved data

pjanik:
[[#184232569]](https://www.pivotaltracker.com/story/show/184232569)

This PR replaces the native browser confirm dialog with a custom one that is consistent with other dialogs that we use in data collection mode. It also adds a few log messages.


pjanik:
@dougmartin, I'll also go ahead and merge this PR to be able to demo everything using master (similarly to this one https://github.com/concord-consortium/tectonic-explorer/pull/313#issuecomment-1383235926). It's pretty straightforward, and I'll address all the comments that you might leave here.

PR #316, opened 2023-01-24
Fix cross-section rotation lock in Data Collection mode

pjanik:
This PR fixes the issue described in the PT story: [[#184239320]](https://www.pivotaltracker.com/story/show/184239320)

It was pretty hard to debug, but apparently, when the same renderer was reused multiple times, OrbitControls were not working correctly anymore. I was disabling rotation correctly, but after the whole CrossSection3D was recreated with an old renderer, setting a flag related to camera rotation in OrbitControls had no longer any effect. 

I tried to create a new renderer each time when the component is created and it works fine. I don't think there's any drawback to this approach. Maybe I was trying to reuse one render for some performance reasons, but I don't think it matters.

### DIFF
--- a/packages/tectonic-explorer/js/get-threejs-renderer.ts
+++ b/packages/tectonic-explorer/js/get-threejs-renderer.ts
@@ -1,6 +1,6 @@
 import * as THREE from "three";
 
-class FakeRenderer {
+export class FakeRenderer {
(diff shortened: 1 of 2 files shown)


dougmartin (approved):
Looks good :+1:

PR #317, opened 2023-01-24
TecRock Table: update layout so snapshots are bigger when possible

pjanik:
[[#184249537]](https://www.pivotaltracker.com/story/show/184249537)


dougmartin (approved):
Looks good :+1:

PR #319, opened 2023-01-25
Tectonic Explorer refactoring: LESS -> SCSS, updated webpack config, js -> src, images -> assets

pjanik:
[[#184314710]](https://www.pivotaltracker.com/story/show/184314710)

Refactoring of Tectonic Explorer source, mostly focused on styles. Changes are split into separate commits, so they should describe well what happened. The goal was to make TE source more similar to other CC projects.

Summary:
1. LESS converted to SASS
2. Styles moved to the components directory, so they live next to related .tsx files (it's way easier for me to navigate between these files that way and that's what we use in other projects)
3. Global styles have .global.scss prefix following the convention set in other CC webpack config 
4. `js` dir renamed to `src`
5. Reorganization of assets - images moved to `src/assets`, and some icons moved there too


packages/tectonic-explorer/src/components/planet-view.global.scss
```diff
@@ -1,18 +1,18 @@
-@import "variables.less";
+@import "./variables.scss";
 
 .planet-view {
   position: relative;
 
   &.full {
-    // ~"..." is a way to escape Less processing. Otherwise, Less compiles it to calc(40%)...
+    // unquote("...") is a way to escape Less processing. Otherwise, Less compiles it to calc(40%)...
```
dougmartin, line 7: Should change to:  `// unquote("...") is a way to escape SASS processing. Otherwise, SASS compiles it to calc(40%)...`

dougmartin (approved):
Wow, lotsof work.  Looks good :+1:  I just found 1 unreplaced Less comment but that shouldn't hold this up from being merged.

packages/tectonic-explorer/src/components/planet-view.global.scss
pjanik: Thanks, I just fixed it in the last commit (and moved to a better place actually).

PR #322, opened 2023-02-12
Update serialized state version, add v4 to v5 migration

pjanik:
[[#184421894]](https://www.pivotaltracker.com/story/show/184421894)

More context: https://concord-consortium.slack.com/archives/C01F450JVE1/p1675707376900469 

Trudi noticed that TE interactives that load serialized models don't work anymore. I accidentally made TE v2.5.0 not backward compatible with v2.4.1. This was not necessary. This PR updates the version number of the serialized state and adds automatic migration.


dougmartin (approved):
Looks good :+1:

PR #324, opened 2023-04-28
Position data collection dialog boxes at top of cross section (PT-185002507)

lublagg:


packages/tectonic-explorer/src/components/data-collection-dialog.tsx
```diff
@@ -44,12 +44,16 @@ export const DataCollectionDialog: React.FC<IProps> = ({ currentDataSample, onCl
   const columnsWithoutIdAndNotes = config.dataSampleColumns.filter((column: DataSampleColumnName) => column !== "notes" && column !== "id");
   const sortedColumns = getSortedColumns(columnsWithoutIdAndNotes);
 
+  // 350 is the height of the cross-section canvas. Place the dialog slightly above the canvas.
+  const yOffset = -350-45;
```
dougmartin, line 48: I think it would be better to put this constant in one file as you repeat it in 3 files.

dougmartin (changes requested):
Can you move the `yOffset` to 1 file (and there are build issues).

PR #325, opened 2023-05-03
Update s3-deploy to use s3-deploy-action

lublagg:


dougmartin (changes requested):
Could you also delete these now unneeded files?

- s3_deploy.sh
- s3_website.yml

PR #328, opened 2023-05-08
Limit pins user can place to collect samples (PT-185002437)

lublagg:


packages/tectonic-explorer/src/plates-interactions/helpers.ts
```diff
@@ -2,11 +2,13 @@ import $ from "jquery";
 import RockSampleCursorSrc from "../assets/rock-sample-cursor.png";
 import TempPressureCursorPng from "../assets/temp-pressure-cursor.png";
 import CollectDataCursorPng from "../assets/collect-data-cursor.png";
+import CollectDataCursorDisabledPng from "../assets/collect-data-cursor-disabled.png";
 
 import { IVector2 } from "../types";
 
 export const TakeRockSampleCursor = `url("${RockSampleCursorSrc}") 16 42, crosshair`;
 export const CollectDataCursor = `url("${CollectDataCursorPng}") 36 35, crosshair`;
+export const CollectDataCursorDisabled = `url("${CollectDataCursorDisabledPng}) 36 35, crosshair`;
```
pjanik, line 11: You're missing closing " :wink:

```suggestion
export const CollectDataCursorDisabled = `url("${CollectDataCursorDisabledPng}") 36 35, crosshair`;
```

packages/tectonic-explorer/src/stores/simulation-store.ts
```diff
@@ -210,6 +216,14 @@ export class SimulationStore {
     return result;
   }
 
+  @computed get arePinsLimited() {
+    return this.limitPins;
+  }
+
+  @computed get reachedPinLimit() {
+    return this.pinLimitReached;
+  }
```
pjanik, line 225: We don't really need this state variable - I think it's redundant. This computed property could check if `dataSamplesLength < 20`. That way you also limit the room for bugs, as this redundant state could get out of sync with the actual number of pins. Not a big deal here, but it's good practice to try to compute props as long as possible. One exception could be when calculations are too expensive and should be cached. Then, 
 this kind of redundant state becomes useful. But here it's only a simple comparison between two numbers.

pjanik (approved):
Looks good 👍

PR #329, opened 2023-05-08
Use smaller metamorphic diagrams in data table. (PT-185002333)

lublagg:


pjanik (approved):
Looks good 👍 
Personally, I'd change .tableDiagram name to something like .metamorphicGradeTableDiagram, so it's more clear what that is and to avoid conflicts if we add more table-specific diagrams/images. But it's very minor.

Edit: I've just noticed that in the subsequent PR you used .table<...> prefix, so I guess it could work here too.

PR #331, opened 2023-05-09
Display placeholder graphic when data is cleared. (PT-185002145)

lublagg:
Before, placeholder graphic was only displaying when the dataSamples array was not defined, which led to the graphic not displaying when collection data was cleared. Now, display the graphic when array is defined but has no length.


pjanik (approved):
Looks good 👍

PR #332, opened 2023-05-25
 Data collection dialog boxes viewable when in Chromebook. (PT-185002507)

lublagg:
This PR addresses the following issues in [this story](https://www.pivotaltracker.com/story/show/185002507):
- Shows full data collection dialog box when sim is viewed in Chromebook (even if it covers cross-section) 
- If not in Chromebook, data collection dialog boxes are pinned to top of the cross-section


pjanik (commented):
The description of this PR suggest that there's some conditional logic that either pins the dialog at the top of the cross-section (when possible), or just places it wherever so it's not cut off.

It works well for the small window sizes, the dialog is never cut off, but I never see the dialog pinned at the top of the cross-section, eg here a large window example:
Screenshot 2023-05-26 at 12 49 36

Or am I misinterpreting something?

pjanik (approved):
@lublagg, as, I see. Sorry, I misinterpreted the description. Maybe it was also caused by Chris' comment here:
https://www.pivotaltracker.com/story/show/185002507/comments/236757135
that also sounds to me like he was expecting some kind of conditional behavior. But that's something to check with PIs, on my side this PR looks good then. 👍

PR #334, opened 2023-07-03
chore: Add release symlink workflow

433eros:


pjanik (approved):
I think it looks good 👍 
It'll release only simulation part (not the data table), but it seems fine. Data Table it's only used as a interactive, so versioned URL makes sense there.

PR #336, opened 2023-12-13
chore(webpack): Fix release process for Tecrocks Table

433eros:
See PT-186673191 for details about issues these changes address

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,10 +15,16 @@ jobs:
   release:
     runs-on: ubuntu-latest
     steps:
-      - run: >-
-          aws s3 cp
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -15,10 +15,16 @@ jobs:
   release:
     runs-on: ubuntu-latest
     steps:
-      - run: >-
-          aws s3 cp
-          s3://${{ env.BUCKET }}/${{ env.PREFIX }}/version/${{ github.event.inputs.version }}/${{ env.SRC_FILE }}
-          s3://${{ env.BUCKET }}/${{ env.PREFIX }}/index.html
+      - run: |-
```
dougmartin, line 18: I think the existing "chomp" (`>-`) method is easier to read than the new version which requires the `\` line continuation character but that is just personal taste.

dougmartin (approved):
Looks good :+1:

PR #337, opened 2024-01-05
Update ThreeJS to r160

pjanik:
https://www.pivotaltracker.com/story/show/186622948

This PR updates ThreeJS to r160. Necessary changes include:
- Updating custom shaders. This is expected, and I hope my improved comments make future modifications easier.
- Updates necessary to address the new color management introduced in ThreeJS r152/153: [ThreeJS Color Management](https://threejs.org/docs/#manual/en/introduction/Color-management). This is a bit painful, as there is no way to convert the code and achieve exactly identical results. Thus, the updates involve setting the correct color space in various places, converting some colors from sRGB to linear sRGB, and adjusting lighting.
- One of the more recent ThreeJS versions also updated the logic related to bump mapping. This required some tweaking of the TecRock code to achieve a similar effect as before.

Generally, the colors, lighting, and bump mapping are not 100% identical, but I hope they are close enough to avoid any noticeable differences.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33937,9 +33937,9 @@
       "dev": true
     },
     "node_modules/three": {
-      "version": "0.151.0",
-      "resolved": "https://registry.npmjs.org/three/-/three-0.151.0.tgz",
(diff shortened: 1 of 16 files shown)


lublagg (approved):
Looks good to me! 👍

PR #338, opened 2024-01-09
Maintenance / dependency update

pjanik:
This PR updates as many dependencies as possible within a reasonable time frame. Note that there is a Cypress bug that causes some tests to fail when using TypeScript v5+ (https://github.com/cypress-io/cypress/issues/26203). Consequently, I relocated some tests out of the basic directory. However, all of these tests were already being skipped. With Cypress v13, the tests that are currently in use now work.

Dependencies in tecrock-table and shared packages have also been updated. I mostly had to skip updating React or React types, as these updates cause issues with the question-interactives-helpers dependency, which we import in an unconventional manner (by directly reading the source from the package).


lublagg (approved):
Looks good! 👍

kswenson:
> Note that there is a Cypress bug that causes some tests to fail when using TypeScript v5+ ([cypress-io/cypress#26203](https://github.com/cypress-io/cypress/issues/26203)).

In CODAP, I created a cypress-specific `tsconfig.json`:
```
{
  "extends": "../tsconfig.json",
  "compilerOptions": {
    "sourceMap": false
  },
  "include": [
    "./**/*"
  ]
}
```

pjanik:
> > Note that there is a Cypress bug that causes some tests to fail when using TypeScript v5+ ([cypress-io/cypress#26203](https://github.com/cypress-io/cypress/issues/26203)).
> 
> In CODAP, I created a cypress-specific `tsconfig.json`:
> 
> ```
> {
>   "extends": "../tsconfig.json",
>   "compilerOptions": {
>     "sourceMap": false
>   },
>   "include": [
>     "./**/*"
>   ]
> }
> ```

Right, I've seen this recommended as a workaround in the Cypress bug thread. I've tried it, but it didn't help in this case. It seems that the main tsconfig.json is being picked up anyway. Moreover, looking at the Cypress bug thread, it appears I'm not the only one experiencing this issue.

To provide more details: this bug occurs only in tests that import code from outside the Cypress directory (like helpers, types). Tests contained within Cypress dir work fine, with or without this local, Cypress-specific tsconfig.json.

PR #339, opened 2024-01-09
Rename packages/tectonic-explorer to packages/tecrock-simulation

pjanik:
It was a bit annoying to me that the top-level directory had the same name as one of the packages. My terminal shows only the last segment of the path, so I never knew exactly where I was (and it's often useful to be in the top-level directory because of Lerna scripts). I'm happy to change this name to something else if you have suggestions. :)


lublagg (approved):
Looks good!

PR #340, opened 2024-01-10
Update the rock pattern rendering so that the pattern follows the position of the cells rather than being stationary

pjanik:
https://www.pivotaltracker.com/story/show/186761171

The PT story should describe the reasons and goals behind this change. The `.setTransform` method has made this change quite concise. Initially, I was concerned that I would need to render patterns manually using rectangles + clipping.
I think it works quite well in practice. Let's see what the project team says.


lublagg (approved):
LGTM 👍

PR #342, opened 2024-01-16
 Add new divergent boundary magma animation

pjanik:
https://www.pivotaltracker.com/story/show/186817402

This PR replaces the previous divergent boundary magma animation with one designed by Ryan. Frames are different, as there's left, right and center parts. Also, there's a new logic that hides left or right frame ("magma arm") when one of the places is not moving:

The project team and/or Ryan wanted to avoid this case:
Screenshot 2024-01-15 at 21 14 40

The current implementation:
Screenshot 2024-01-16 at 23 08 28

The approach I've taken is the simplest one I could find. While I'm certain it doesn't cover 100% of the cases, I hope it should be preferable to implementing overly complicated logic for scenarios that are unlikely to be relevant.


lublagg (approved):
Nice, looks good! 👍

PR #343, opened 2024-01-18
Fix a bug causing a red gradient to appear around continental divergent boundaries

pjanik:
This PR addresses a bug described here:
https://www.pivotaltracker.com/story/show/186640790/comments/239865995
I commented out some logic earlier while experimenting with this gradient (initial work on this story).


lublagg (approved):
LGTM 👍

PR #347, opened 2024-01-25
Implement tecrock-table report item, render table to string, inline CSS

pjanik:
https://www.pivotaltracker.com/story/show/186622903

This PR introduces a report item to the tecrock table. 

So far, the report item renders the Table component (extracted from the tecrock-table runtime) into a string and inlines the CSS styles. To accomplish this, I am loading the SCSS file as a raw string using a Webpack loader (the SASS still requires processing). Then, I replace the class names with those set by the CSS modules. Finally, I use the `juice` library to merge HTML with CSS and generate HTML with inline styles.

This approach seems to function well within the Dashboard. The Lato font is not rendered, but I consider that to be a minor issue at this stage.

The next step is to update the layout of the report item to incorporate Trudi's suggestions.

I've also started the `report-item` branch to avoid merging all these changes into master, as we're currently in the research spike stage.


packages/tecrock-table/src/components/runtime.scss
```diff
@@ -21,70 +21,7 @@
...
-          bottom: 3px;
-          right: 3px;
-          height: 30px;
-          pointer-events: none;
-        }
-      }
-    }
-  }
-
+  // TODO: remove?
```
lublagg, line 24: Do you want to remove this comment?

lublagg (approved):
Looks good, just one question about a commented-out line.

packages/tecrock-table/src/components/runtime.scss
pjanik: Right, thanks. Fixed in the last commit.

PR #348, opened 2024-01-26
Scale down the table in the Question Details view, allow teachers to enlarge snapshots

pjanik:
https://www.pivotaltracker.com/story/show/186921649
https://www.pivotaltracker.com/story/show/186921720

This PR scales down the table in the Question Details view and allows teachers to enlarge snapshots. It's done using a pretty basic JavaScript script, as the report-item doesn't have access to any libraries. Fortunately, in this case, the basic DOM API was sufficient.


lublagg (approved):
LGTM 👍

PR #351, opened 2024-02-01
Fix report-item layout script

pjanik:
This PR addresses the issue that Saravanan described in the second bullet point of his comment, which can be found here: https://www.pivotaltracker.com/story/show/186921649/comments/240062521

The issue is actually triggered by the dashboard animation of the report item container.


pjanik:
@lublagg, the project team wanted to release TecRock today, so I'm going to merge this PR as Saravanan confirmed that the fix works. If you have any comments, please leave them here anyway, and I'll address them.

PR #352, opened 2024-02-12
Fix index.html template

pjanik:
https://www.pivotaltracker.com/story/show/186957289

This doesn't seem related to the bug, but I believe the issue is that the script tag was added twice: once via HtmlWebpackPlugin and then again through an explicit `<script src="app.js"></script>` line. This wasn't visible in most cases because the same script was loaded twice, until we released this project to production. Two slightly different scripts were loaded (from `version/v2.7.0` and top-level production dir), which possibly caused the linked bug.

### DIFF
--- a/packages/tecrock-simulation/src/index.html
+++ b/packages/tecrock-simulation/src/index.html
@@ -40,6 +40,5 @@
 </head>
 <body>
 <div id="app"></div>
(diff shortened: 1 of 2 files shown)


dougmartin (approved):
Looks good 👍

PR #353, opened 2024-02-12
Fix release process, handle data assets correctly, rm old favicons

pjanik:
https://www.pivotaltracker.com/story/show/187028887

This PR removes the CopyWebpackPlugin, as this approach has never worked correctly with the new release mechanism. Things worked fine until we tried to update one of these images. So, from now on, the preset images are loaded using webpack. I've also removed some old favicon configurations and metadata, as it doesn't seem we set them that way in any other project. If we want to be really particular about favicons, there are webpack plugins that can generate everything for us.


dougmartin (approved):
Looks good 👍

PR #354, opened 2024-02-12
Load Web Worker script from a correct location

pjanik:
https://www.pivotaltracker.com/story/show/187028887
https://www.pivotaltracker.com/story/show/186957289

Another PR has been created to improve the new release process. The Web Worker script was being loaded from a local directory instead of the /version/v2.7.0 directory after the production release and that was causing the issue reported by Trudi.

I've explored better approaches, and there are two separate ones:
- The recommended approach by Webpack 5, as detailed at https://webpack.js.org/guides/web-workers/, won't work because the project does not use ESM.
- The worker-loader might work, but I was unable to pass custom URL parameters to the worker script to set up the configuration. Using postMessage is not feasible, as a significant amount of code requires certain values to be set at initialization.

So, I have opted to pass the deployment path to the TS script and use it there.

### DIFF
--- a/packages/tecrock-simulation/webpack.config.js
+++ b/packages/tecrock-simulation/webpack.config.js
@@ -176,7 +176,11 @@ module.exports = {
       template: './src/index.html',
       favicon: './public/favicon.ico',
       publicPath: DEPLOY_PATH,
-    })] : [])
+    })] : []),
+    // Define DEPLOY_PATH globally so it can be used to load Web Worker script from a correct location
+    new webpack.DefinePlugin({
+      '__WEBPACK_DEPLOY_PATH__': JSON.stringify(DEPLOY_PATH || ".")
+    })
   ]
 }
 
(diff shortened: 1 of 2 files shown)


dougmartin (approved):
Looks good 👍

packages/tecrock-simulation/webpack.config.js
```diff
@@ -176,7 +176,11 @@ module.exports = {
       template: './src/index.html',
       favicon: './public/favicon.ico',
       publicPath: DEPLOY_PATH,
-    })] : [])
+    })] : []),
+    // Define DEPLOY_PATH globally so it can be used to load Web Worker script from a correct location
+    new webpack.DefinePlugin({
+      '__WEBPACK_DEPLOY_PATH__': JSON.stringify(DEPLOY_PATH || ".")
+    })
```
scytacki, line 183: @pjanik rather than doing this, perhaps you can use the built in webpack variable `__webpack_public_path__`? This is described here: https://github.com/concord-consortium/starter-projects/blob/master/doc/deploy.md

pjanik: Thanks @scytacki, I've made this change in https://github.com/concord-consortium/tectonic-explorer/pull/355.

PR #355, opened 2024-02-13
Use built-in __webpack_public_path__ instead of custom global variable

pjanik:
This PR addresses @scytacki's comment:
https://github.com/concord-consortium/tectonic-explorer/pull/354#discussion_r1487063479
It simplifies the configuration slightly.

I tested it both using webpack-dev server and on production (releasing v2.7.2-pre.1) and it seems to work fine.


scytacki (approved):
👍 Nice.

PR #362, opened 2024-10-08
feat: Updated authoring form [PT-188187229]

dougmartin:
- Changed prompt to be rich text
- Added rich text hint field
- Added check data boolean
- Added list of required rock types derived from a newly exported list of rock types already defined in the shared code


bgoldowsky (approved):
These changes look reasonable.

PR #363, opened 2024-10-08
feat: Added check data button with result message [PT-188187248]

dougmartin:
Shows a "Check Data" button based on the authored state that when checked stores the hash of the current interactive state of the linked interactive.  When the current linked interactive state and the check state hashes match then a check is done to see if the user selected all the required rocks set in the interactive state.


bgoldowsky (approved):
We did a synch review on this, looks good.

PR #365, opened 2025-03-13
feat: Change save/load of models to use S3 [TE-2]

dougmartin:
This is phase one of the migration of the models to S3.  In this phase new models are saved to S3 using the Token Service and loaded from S3 with a fallback to loading from Firebase when they are not found on S3.

Test links:

- New model saved to S3: https://tectonic-explorer.concord.org/branch/move-models-to-s3/index.html?modelId=Q316ZpnzqaNUeG24tvRD&rocks=false
- Old model manually migrated to S3: https://tectonic-explorer.concord.org/branch/move-models-to-s3/index.html?modelId=08971159-a0b4-4b02-aead-29719faf4e64&rocks=false
- Old model not migrated to S3, will fallback to Firebase: https://tectonic-explorer.concord.org/branch/move-models-to-s3/index.html?modelId=d9ac2f07-d093-4b5e-8f14-2a8330921614&rocks=false

### DIFF
--- a/packages/tecrock-simulation/src/storage.ts
+++ b/packages/tecrock-simulation/src/storage.ts
@@ -1,7 +1,9 @@
+import { TokenServiceClient, S3Resource } from "@concord-consortium/token-service";
 import { FirebaseApp, initializeApp } from "firebase/app";
-import { set, ref, onValue, getDatabase } from "firebase/database";
+import { ref, onValue, getDatabase } from "firebase/database";
+import pako from "pako";
+import S3 from "aws-sdk/clients/s3";
 import { ISerializedState } from "./stores/simulation-store";
-import { v4 as uuidv4 } from "uuid";
 
 let app: FirebaseApp | null = null;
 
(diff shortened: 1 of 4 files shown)


packages/tecrock-simulation/src/stores/simulation-store.ts
```diff
@@ -429,6 +429,9 @@ export class SimulationStore {
   @action.bound loadCloudModel(modelId: string) {
     this.modelState = "loading";
     loadModelFromCloud(modelId, (serializedModel: ISerializedState) => {
+      if (!serializedModel) {
```
scytacki, line 432: If `serializeModel` can be undefined or null, the type of loadModelFromCloud should be changed too.

packages/tecrock-simulation/src/storage.ts
```diff
@@ -16,33 +18,106 @@ export function initDatabase() {
...
+      await s3.upload({
+        Bucket: bucket,
+        Key: publicPath,
+        Body: blob,
+        ContentType: "application/json",
+        ContentEncoding: "gzip",
+        CacheControl: "public, max-age=31536000, immutable", // Cache forever (1 year)
+      }).promise();
+
+      const [_folder, id, _filename] = publicPath.split("/");
```
scytacki, line 117: Looks like you could get the id directly with `resource.id` instead of parsing the path which might change via a configuration change within the token service.

scytacki (approved):
Looks good. Just a couple of suggestions.

packages/tecrock-simulation/src/storage.ts
dougmartin: Good point.  I'll update the code.

dougmartin: Done in https://github.com/concord-consortium/tectonic-explorer/pull/365/commits/b1c10dc9838b17cd21b9943eecdd3984fa4edbe6

packages/tecrock-simulation/src/stores/simulation-store.ts
dougmartin: Done in https://github.com/concord-consortium/tectonic-explorer/pull/365/commits/b1c10dc9838b17cd21b9943eecdd3984fa4edbe6

PR #367, opened 2025-03-21
feat: Remove model load fallback to Firebase [TE-5]

dougmartin:
All models are now loaded/stored on S3.


scytacki (approved):
Looks good to me. 👍

